PR #21, opened 2022-03-24
Added help chart uninstall action handler

andrejatcastai:


actions/chart_uninstall_handler_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+
+	"github.com/castai/cluster-controller/castai"
+	"github.com/castai/cluster-controller/helm"
+	mock_helm "github.com/castai/cluster-controller/helm/mock"
+)
+
+func TestUninstallActionValidator(t *testing.T) {
```
anjmao, line 17: It's not a good practice to test private methods. In this case you can test it via `TestChartUninstallHandler` subtest.

anjmao (approved):

actions/chart_uninstall_handler_test.go
andrejatcastai: Changed as suggested

PR #29, opened 2022-04-19
aks init data rework

anjmao:


main.go
```diff
@@ -58,6 +59,22 @@ func main() {
...
+	// Send aks init data and exit if run from temporary provisioned node.
+	if cfg.AksInitData {
+		h := aks.NewInitDataHandler(log, client)
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+		defer cancel()
+		if err := h.Handle(ctx); err != nil {
+			log.Fatalf("sending aks init data: %v", err)
+		}
+		log.Info("aks init data sent, waiting for init data node shutdown")
+		time.Sleep(time.Hour)
```
varnastadeus, line 75: do we really need this?

varnastadeus (approved):

main.go
anjmao: If pod exists it will be scheduled again. Instead we can wait until node is removed.

PR #30, opened 2022-04-25
Helm should ignore some manifests if only version changed

matas-cast:
…art manifests


helm/client.go
```diff
@@ -3,26 +3,42 @@
...
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
+	"sigs.k8s.io/yaml"
 
 	"github.com/castai/cluster-controller/castai"
 )
 
+// group/version/kind/namespace/name
+var ignoredCharts = map[string]struct{}{
```
anjmao, line 32: Maybe name it something like `labelIgnoreResources`

helm/client.go
```diff
@@ -3,26 +3,42 @@
...
+	"rbac.authorization.k8s.io/v1/ClusterRole//castai-evictor":        {},
+	"rbac.authorization.k8s.io/v1/ClusterRoleBinding//castai-evictor": {},
+}
+
+const (
+	K8sVersionLabel  = "app.kubernetes.io/version"
```
anjmao, line 38: Make these private if not used in other packages

helm/client.go
```diff
@@ -80,6 +96,70 @@ type client struct {
 	chartLoader         ChartLoader
 }
 
+// LabelIgnoreHook prevents certain charts getting updated, if only their version labels have changed
```
anjmao, line 99: Maybe
```
// LabelIgnoreHook prevents certain resource getting updated, if only their version labels have changed.
// This is needed in order to update components like evictor with it's own cluster scoped resources like clusterrole.
```

anjmao (approved):

helm/hook/hook.go
```diff
@@ -0,0 +1,95 @@
...
+func getChartLabels(list kube.ResourceList, chartName, kind, namespace string) map[string]string {
+	for _, r := range list {
+		u := r.Object.(*unstructured.Unstructured)
+		if u.GetName() == chartName && u.GetKind() == kind && u.GetNamespace() == namespace {
+			return u.GetLabels()
+		}
+	}
+
+	return nil
+}
```
anjmao, line 95: Add new lines

anjmao (approved):

PR #33, opened 2022-05-19
feat: node patch improvements

anjmao:


actions/patch_node_handler.go
```diff
@@ -30,6 +31,21 @@ func (h *patchNodeHandler) Handle(ctx context.Context, data interface{}) error {
 	if !ok {
 		return fmt.Errorf("unexpected type %T for delete patch handler", data)
 	}
+	for k := range req.Labels {
+		if k == "" {
+			return errors.New("labels contain entry with empty key")
```
varnastadeus, line 36: do we have similar validation before scheduling the action?

varnastadeus, line 36: do we have similar validation before scheduling the action?

anjmao: Will add validation, thanks for suggestion.

varnastadeus (approved):

PR #34, opened 2022-05-23
fix: do not ack actions for context canceled errors

anjmao:


actions/drain_node_handler.go
```diff
@@ -277,7 +274,16 @@ func (h *drainNodeHandler) evictPod(ctx context.Context, pod v1.Pod) error {
...
-	return controller != nil && controller.Kind == "DaemonSet"
+func isDaemonSetPod(p *v1.Pod) bool {
+	return isControllerOf(p, "DaemonSet")
+}
+
+func isStaticPod(p *v1.Pod) bool {
+	return isControllerOf(p, "Node")
+}
+
+func isControllerOf(p *v1.Pod, kind string) bool {
```
matas-cast, line 285: random nitpic: feels like this should be `isControlledBy` ?

anjmao: Fixed, isControlledBy is correct method here, thanks.

matas-cast (approved):

varnastadeus (approved):

PR #36, opened 2022-06-07
feat: improve api request count for node drain action

r0kas:
Improve drain node action by:
- batching pod requests in chunks of 5 to not overload control plane api
- do not call for each pod individually when waiting for them to be terminated on the node

### DIFF
--- a/actions/drain_node_handler.go
+++ b/actions/drain_node_handler.go
@@ -20,10 +20,6 @@ import (
 	"github.com/castai/cluster-controller/castai"
 )
 
-var (
-	errPodPresent = errors.New("pod is still present")
-)
-
 type drainNodeConfig struct {
 	podsDeleteTimeout   time.Duration
 	podDeleteRetries    uint64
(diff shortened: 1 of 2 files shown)


actions/drain_node_handler.go
```diff
@@ -183,58 +167,28 @@ func (h *drainNodeHandler) listNodePods(ctx context.Context, node *v1.Node) (*v1
...
-	}, b)
-	if err != nil && errors.Is(err, errPodPresent) {
-		log.Infof("timeout waiting for pod %s in namespace %s to terminate", pod.Name, pod.Namespace)
-		return nil
-	}
-	if err != nil {
-		return fmt.Errorf("waiting for pod %s in namespace %s termination: %w", pod.Name, pod.Namespace, err)
-	}
-	return nil
+	}, backoff.WithContext(backoff.NewConstantBackOff(10*time.Second), ctx))
```
anjmao, line 191: Maybe move this hardcoded duration to drainNodeConfig which contains more of these constants.

actions/drain_node_handler.go
```diff
@@ -107,74 +94,71 @@ func (h *drainNodeHandler) Handle(ctx context.Context, data interface{}) error {
...
-		n.Spec.Unschedulable = true
-		return nil
-	})
-	if err != nil {
-		return fmt.Errorf("patching node unschedulable: %w", err)
+	for _, batch := range lo.Chunk(pods, batchSize) {
+		g, ctx := errgroup.WithContext(ctx)
+		for _, pod := range batch {
+			pod := pod
+			g.Go(func() error { return f(ctx, pod) })
```
anjmao, line 149: Before with single request for pod we were retrying request with backoff. Would be good to wrap f(ctx, pod) into retry here to.

r0kas: Requests are wrapped inside the functions with different params for different call. Check EvictPod, DeletePod methods.

anjmao: Ah, ok, cool!

anjmao (approved):

PR #40, opened 2022-09-08
feat: node status check action

aldor007:
requires new RBAC perm - watch 
https://github.com/castai/helm-charts/pull/143

### DIFF
--- a/actions/check_node_status.go
+++ b/actions/check_node_status.go
@@ -32,7 +32,7 @@ type checkNodeStatusHandler struct {
 func (h *checkNodeStatusHandler) Handle(ctx context.Context, data interface{}) error {
 	req, ok := data.(*castai.ActionCheckNodeStatus)
 	if !ok {
-		return fmt.Errorf("unexpected type %T for check node deleted handler", data)
+		return fmt.Errorf("unexpected type %T for check node status handler", data)
 	}
 
 	log := h.log.WithFields(logrus.Fields{
--- a/actions/check_node_status.go
+++ b/actions/check_node_status.go
@@ -2,16 +2,14 @@ package actions
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"math"
 	"time"
 
-	"github.com/cenkalti/backoff/v4"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	watchv1 "k8s.io/apimachinery/pkg/watch"
+
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/castai/cluster-controller/castai"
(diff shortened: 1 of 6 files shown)


actions/check_node_status.go
```diff
@@ -0,0 +1,88 @@
...
+
+type checkNodeStatusHandler struct {
+	log       logrus.FieldLogger
+	clientset kubernetes.Interface
+}
+
+func (h *checkNodeStatusHandler) Handle(ctx context.Context, data interface{}) error {
+	req, ok := data.(*castai.ActionCheckNodeStatus)
+	if !ok {
+		return fmt.Errorf("unexpected type %T for check node deleted handler", data)
```
varnastadeus, line 35:
```suggestion
		return fmt.Errorf("unexpected type %T for check node status handler", data)
```

actions/check_node_status.go
```diff
@@ -0,0 +1,88 @@
...
+		}, b)
+	} else if req.NodeStatus == castai.ActionCheckNodeStatus_READY {
+		// 10 minutes default timeout
+		timeout := int32(600)
+		if req.WaitTimeoutSeconds != nil {
+			timeout = *req.WaitTimeoutSeconds
+		}
+		retries := uint64(math.Round(float64(timeout / 30)))
+		log.Infof("checking if node is ready retries: %d", retries)
+		b := backoff.WithContext(backoff.WithMaxRetries(backoff.NewConstantBackOff(time.Second*30), retries), ctx)
```
varnastadeus, line 70: maybe 30sec is too long, we will have 30sec delay from node becoming `ready` and us actually noticing it

aldor007: WDYT about 10sec?

actions/check_node_status_test.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+
+	"github.com/castai/cluster-controller/castai"
+)
+
+func TestCheckStatus_Deleted(t *testing.T) {
+	r := require.New(t)
```
varnastadeus, line 19: move this as part of each test

actions/check_node_status.go
```diff
@@ -0,0 +1,127 @@
...
+				}
+			case <-ctx.Done():
+				watch.Stop()
+				return errWatchTimeout
+			}
+		}
+	})
+
+	err = errGroup.Wait()
+	if err == errWatchTimeout {
```
varnastadeus, line 91: why do we need fallback?

aldor007: * it was easier to write unit tests
* on watch timeout we can get real result from the cluster instead of timeout error

actions/check_node_status_test.go
aldor007: Done

actions/check_node_status.go
varnastadeus: - You can mock clientset and return channel to test it
- not sure I got this reason

aldor007: @varnastadeus 
In the case of timeout for node delete action based on timeout is hard to tell if the node was removed or not. For example, if the node wasn't present in the cluster we will get a timeout for events and mark action as failed because of timeout. But the node is not in the cluster. 

I can agree that fallback is not required for checking if a node is ready but added this check as a safeguard if cluster-controller misses an event for the node.

varnastadeus: in that case maybe we should use watch only for `ActionCheckNodeStatus_READY` and for  `ActionCheckNodeStatus_DELETED` use the simple `Nodes().Get` as we don't need to poll the API server to check it

aldor007: @varnastadeus changed as suggested

actions/check_node_status.go
```diff
@@ -0,0 +1,120 @@
...
+
+	timeoutInSec := int64(timeout.Seconds())
+	watchObject.TimeoutSeconds = &timeoutInSec
+
+	watch, err := h.clientset.CoreV1().Nodes().Watch(ctx, watchObject)
+	if err != nil {
+		return fmt.Errorf("creating node watch: %w", err)
+	}
+
+	errGroup, ctx := errgroup.WithContext(ctx)
```
varnastadeus, line 92: `errgroup` is not necessary here

actions/check_node_status.go
```diff
@@ -0,0 +1,120 @@
...
+	timeout := 10 * time.Second
+	watchObject := metav1.SingleObject(metav1.ObjectMeta{Name: req.NodeName})
+	if req.WaitTimeoutSeconds != nil {
+		timeout = time.Duration(*req.WaitTimeoutSeconds) * time.Second
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	timeoutInSec := int64(timeout.Seconds())
+	watchObject.TimeoutSeconds = &timeoutInSec
```
varnastadeus, line 85: why this is needed? we already have `context.WithTimeout(context.Background(), timeout)` so the watch should timeout anyway and setting multiple timeouts adds confussion

actions/check_node_status.go
```diff
@@ -0,0 +1,120 @@
...
+					}
+				}
+			case <-ctx.Done():
+				watch.Stop()
+				return fmt.Errorf("timeout waiting for node %s to become ready", req.NodeName)
+			}
+		}
+	})
+
+	return errGroup.Wait()
```
varnastadeus, line 109: you can simplify this, eg:
```go
for r := range watch.ResultChan() {
		if node, ok := r.Object.(*corev1.Node); ok {
			if isNodeReady(node.Status.Conditions) {
				return nil
			}
		}
	}

	return fmt.Errorf("timeout execeed waiting for node %q to be ready", w.name)
	```

File: actions/check_node_status.go [hunk 8]
```diff
@@ -0,0 +1,114 @@
...
+
+	}
+
+	return fmt.Errorf("unknown status to check provided node=%s status=%s", req.NodeName, req.NodeStatus)
+}
+
+func (h *checkNodeStatusHandler) checkNodeDeleted(ctx context.Context, req *castai.ActionCheckNodeStatus) error {
+	retries := uint64(5)
+	if req.WaitTimeoutSeconds != nil {
+		retries = uint64(*req.WaitTimeoutSeconds)
```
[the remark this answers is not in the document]
aldor007: What you are suggesting, then? Currently, there is 1s step delay between API call attempts, so for 5s timeout we will have 5 retries.

actions/check_node_status.go
aldor007: Changed as suggested

varnastadeus (approved):

PR #41, opened 2022-09-20
feat: add action health check

aldor007:


actions/healthz_test.go
```diff
@@ -0,0 +1,70 @@
+package actions
```
varnastadeus, line 1: move it to `health` pkg? 🤔

varnastadeus (approved):

PR #43, opened 2022-09-27
feat: drain add fallback to remove pods with grace period 0

aldor007:


go.mod
```diff
@@ -14,11 +14,11 @@ require (
 	go.uber.org/goleak v1.1.12
 	golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f
 	helm.sh/helm/v3 v3.9.2
-	k8s.io/api v0.24.2
-	k8s.io/apimachinery v0.24.2
+	k8s.io/api v0.25.2
```
aldor007, line 17: Contains fix for testing lib https://github.com/kubernetes/kubernetes/pull/110425

actions/drain_node_handler.go
```diff
@@ -54,7 +54,7 @@ func (h *drainNodeHandler) Handle(ctx context.Context, data interface{}) error {
 		return fmt.Errorf("unexpected type %T for delete drain handler", data)
 	}
 
-	log := h.log.WithField("node_name", req.NodeName)
```
andrejatcastai, line 57: This is a drain handler, please consider removing 'delete' from the logging message above.

actions/drain_node_handler.go
```diff
@@ -119,36 +130,40 @@ func (h *drainNodeHandler) evictNodePods(ctx context.Context, log logrus.FieldLo
...
-	if err := h.sendPodsRequests(ctx, pods, h.evictPod); err != nil {
+	if err := h.sendPodsRequests(ctx, metav1.DeleteOptions{}, pods, h.evictPod); err != nil {
 		return fmt.Errorf("sending evict pods requests: %w", err)
 	}
 
 	return h.waitNodePodsTerminated(ctx, node)
 }
 
-func (h *drainNodeHandler) deleteNodePods(ctx context.Context, log logrus.FieldLogger, node *v1.Node) error {
+func (h *drainNodeHandler) deleteNodePods(ctx context.Context, log logrus.FieldLogger, options metav1.DeleteOptions, node *v1.Node) error {
```
andrejatcastai, line 140: It is a common practice to have options as the last parameter in a signature.

andrejatcastai, line 140: I wonder have you considered using functional options pattern? https://golang.cafe/blog/golang-functional-options-pattern.html

actions/drain_node_handler.go
```diff
@@ -195,7 +210,7 @@ func (h *drainNodeHandler) waitNodePodsTerminated(ctx context.Context, node *v1.
 
 // evictPod from the k8s node. Error handling is based on eviction api documentation:
 // https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#the-eviction-api
-func (h *drainNodeHandler) evictPod(ctx context.Context, pod v1.Pod) error {
+func (h *drainNodeHandler) evictPod(ctx context.Context, options metav1.DeleteOptions, pod v1.Pod) error {
```
varnastadeus, line 213: evict doesn't not need `DeleteOptions` so I would suggest keeping method signature as it was
and while deleting pods do:

```go

fn := func(ctx context.Context, pod v1.Pod) error {
	return h.deletePod(ctx, options, pod)
}

if err := h.sendPodsRequests(ctx, pods, fn); err != nil {
	return fmt.Errorf("sending delete pods requests: %w", err)
}


```

actions/drain_node_handler.go
```diff
@@ -86,9 +86,20 @@ func (h *drainNodeHandler) Handle(ctx context.Context, data interface{}) error {
...
+		}
+		if !errors.Is(err, context.DeadlineExceeded) {
 			return fmt.Errorf("forcefully deleting pods: %w", err)
 		}
+
+		deleteForceCtx, deleteForceCancel := context.WithTimeout(ctx, h.cfg.podsDeleteTimeout)
+		defer deleteForceCancel()
+		if err = h.deleteNodePods(deleteForceCtx, log, node, *metav1.NewDeleteOptions(0)); err != nil {
+			return fmt.Errorf("deleting gracePeriod=0 pods: %w", err)
+		}
```
varnastadeus, line 102: suggestion, maybe this will be more clear: 🤔 
```go

// Try deleteing pods gracefully first, then delete with 0 grace period.
		options := []metav1.DeleteOptions{
			metav1.DeleteOptions{},
			*metav1.NewDeleteOptions(0)
		}

		var err error
		for _, o := options {
			deleteCtx, deleteCancel := context.WithTimeout(ctx, h.cfg.podsDeleteTimeout)
			defer deleteCancel()

			err = h.deleteNodePods(deleteCtx, log, node, o);
			if err == nil {
				return nil
			}
			if !errors.Is(err, context.DeadlineExceeded) {
				return fmt.Errorf("forcefully deleting pods: %w", err)
			}
		}

		return err
```

varnastadeus (approved):

PR #45, opened 2022-10-06
feat: improve csr handling

varnastadeus:


actions/approve_csr_handler.go
```diff
@@ -100,17 +95,19 @@ func (h *approveCSRHandler) handle(ctx context.Context, log logrus.FieldLogger,
 	return errors.New("certificate signing request was not approved")
 }
 
-func (h *approveCSRHandler) getInitialNodeCSR(ctx context.Context, nodeName string) (*csr.Certificate, error) {
-	csrFetchCtx, cancel := context.WithTimeout(ctx, h.initialCSRFetchTimeout)
+func (h *approveCSRHandler) getInitialNodeCSR(ctx context.Context, log *logrus.Entry, nodeName string) (*csr.Certificate, error) {
+	log.Debug("getting initial csr")
+
+	ctx, cancel := context.WithTimeout(ctx, h.initialCSRFetchTimeout)
```
fals, line 101: question, we are basically replacing the original ctx timeout with this timeout, won't that cascade upstream? Shouldn't we have a single time out for this whole execution or maybe create a new context and not cancel the main context because of this failure?

varnastadeus: we are not replacing it, it will cascade downstream with this timeout

fals (approved):
LGTM

csr/csr.go
```diff
@@ -244,7 +259,12 @@ func getNodeCSRV1Beta1(ctx context.Context, client kubernetes.Interface, nodeNam
 
 	for _, item := range csrList.Items {
```
aldor007, line 260: Why we aren't using a watch for CSR like in V1 API?

actions/approve_csr_handler.go
```diff
@@ -100,17 +95,19 @@ func (h *approveCSRHandler) handle(ctx context.Context, log logrus.FieldLogger,
...
 	for {
 		select {
-		case <-csrFetchCtx.Done():
-			return nil, csrFetchCtx.Err()
+		case <-ctx.Done():
+			return nil, ctx.Err()
 		case <-time.After(h.csrFetchInterval):
 			cert, err := csr.GetCertificateByNodeName(ctx, h.clientset, nodeName)
-			if err != nil && !errors.Is(err, csr.ErrNodeCertificateNotFound) {
+			if err != nil && errors.Is(err, context.DeadlineExceeded) {
```
aldor007, line 110:
```suggestion
			if errors.Is(err, context.DeadlineExceeded) {
```

Do we want to execit only on context deadline exceeded error?

varnastadeus: changed to retry other errors only few times

csr/csr.go
varnastadeus: the beta path is legacy and we don't have any clusters using it, didn't want to invest much time into it and rather remove it completely in the future

aldor007 (approved):

PR #46, opened 2022-10-07
fix: use polling while handling csr

varnastadeus:


Narunas-K (approved):

actions/approve_csr_handler.go
```diff
@@ -54,21 +61,27 @@ func (h *approveCSRHandler) Handle(ctx context.Context, data interface{}) error
...
 	if err := csr.DeleteCertificate(ctx, h.clientset, cert); err != nil {
 		return fmt.Errorf("deleting csr: %w", err)
 	}
 
+	var newCert *csr.Certificate
+	var err error
+
+	defer func() {
+		if errors.Is(reterr, context.Canceled) && newCert == nil {
+			// Rollback csr deletion if cluster-controller is restarted.
```
aldor007, line 76: Can you explain here more why we are requesting a certificate once more?

varnastadeus: if cluster controller gets restarted in the middle of approval, we will lose csr forever

varnastadeus: tested this again and it looks like kubelet keeps requesting csr if it gets deleted so we don't need this rollback

PR #47, opened 2022-10-13
feat: Improve logging

Narunas-K:


aldor007 (approved):

castai/types.go
```diff
@@ -75,6 +75,28 @@ func (c *ClusterAction) Data() interface{} {
...
+	if c.ActionApproveCSR != nil {
+		return c.ActionApproveCSR.NodeID
+	}
+	if c.ActionCheckNodeDeleted != nil {
+		return c.ActionCheckNodeDeleted.NodeID
+	}
+	if c.ActionCheckNodeStatus != nil {
+		return c.ActionCheckNodeStatus.NodeID
+	}
+	return "nil"
```
andrejatcastai, line 97: From action API perspective it looks like NodeID is optional (for example it is irrelevant for AddNode action). Then return value should be pointer. It means that there should IFs where node id is used for logging.

Narunas-K: Yeah that makes sense. I have removed this part completely as having node ID in the handleAction and ackAction methods looks excessive.

actions/approve_csr_handler.go
```diff
@@ -30,13 +31,17 @@ type approveCSRHandler struct {
 	csrFetchInterval       time.Duration
 }
 
-func (h *approveCSRHandler) Handle(ctx context.Context, data interface{}) error {
+func (h *approveCSRHandler) Handle(ctx context.Context, data interface{}, actionID string) error {
```
varnastadeus, line 34: maybe instead of passing actionID, we should just pass whole action and then use `action.Data()` for actual action data

varnastadeus (approved):

PR #48, opened 2022-10-31
feat: delete all pods when node is deleted

aldor007:


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+*  @castai/cast-core
```
r0kas, line 1: No new line at end of file

actions/delete_node_handler.go
```diff
@@ -52,12 +61,48 @@ func (h *deleteNodeHandler) Handle(ctx context.Context, action *castai.ClusterAc
...
+	pods, err := h.clientset.CoreV1().Pods(metav1.NamespaceAll).List(ctx, metav1.ListOptions{
+		FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": req.NodeName}).String(),
+	})
+	if err != nil {
+		return fmt.Errorf("listing node pods %w", err)
+	}
+
+	log.Infof("node has %d pods - removing", len(pods.Items))
+
+	deleteOptions := metav1.NewDeleteOptions(0)
```
r0kas, line 86: What does '0' mean in these delete opts?

actions/delete_node_handler.go
```diff
@@ -52,12 +61,48 @@ func (h *deleteNodeHandler) Handle(ctx context.Context, action *castai.ClusterAc
...
+		return fmt.Errorf("sending delete pods requests: %w", err)
+	}
+
+	podsWait := backoff.WithContext(backoff.WithMaxRetries(backoff.NewConstantBackOff(h.cfg.podsTerminationWait), h.cfg.deleteRetries), ctx)
+	return backoff.Retry(func() error {
+		pods, err := h.clientset.CoreV1().Pods(metav1.NamespaceAll).List(ctx, metav1.ListOptions{
+			FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": req.NodeName}).String(),
+		})
+		if err != nil {
+			return fmt.Errorf("waiting for node %q pods to be terminated: %w", req.NodeName, err)
```
r0kas, line 101: This error is about listing the pods, would modify the error message to "listing node %q pods"

actions/delete_node_handler.go
```diff
@@ -52,12 +61,48 @@ func (h *deleteNodeHandler) Handle(ctx context.Context, action *castai.ClusterAc
...
+	err := backoff.Retry(func() error {
 		err := h.clientset.CoreV1().Nodes().Delete(ctx, req.NodeName, metav1.DeleteOptions{})
 		if apierrors.IsNotFound(err) {
 			log.Info("node not found, skipping delete")
 			return nil
 		}
 		return err
 	}, b)
+
+	if err != nil {
```
r0kas, line 73: Why not wrap this if around the backoff function. And add some context to returned error message.

actions/delete_node_handler.go
```diff
@@ -52,12 +61,48 @@ func (h *deleteNodeHandler) Handle(ctx context.Context, action *castai.ClusterAc
...
+	deleteOptions := metav1.NewDeleteOptions(0)
+	deletePod := func(ctx context.Context, pod v1.Pod) error {
+		return h.deletePod(ctx, *deleteOptions, pod)
+	}
+
+	if err := h.sendPodsRequests(ctx, pods.Items, deletePod); err != nil {
+		return fmt.Errorf("sending delete pods requests: %w", err)
+	}
+
+	podsWait := backoff.WithContext(backoff.WithMaxRetries(backoff.NewConstantBackOff(h.cfg.podsTerminationWait), h.cfg.deleteRetries), ctx)
```
r0kas, line 95: 25 seconds is enough for this operation?
Would be great to have these options configurable from env. We would have more flexibility if we would need to adjust the timeouts.

r0kas (changes requested):

actions/delete_node_handler.go
aldor007: grace period 0

r0kas: why not init a var to name that? Easier to read

actions/delete_node_handler.go
aldor007: Even 5s is enough for this operation, the node on which those pods are running is gone already.

We don't have env configuration options for any of  actions.  All actions have static config. I can introduce config from env but IMO this would be out of this PR.

actions/delete_node_handler.go
aldor007: Added context.

actions/delete_node_handler.go
aldor007: Added comment is enough or should I rename var?

r0kas (approved):

PR #49, opened 2022-11-02
feat: arm64 support

aldor007:


Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM alpine:3.13
 RUN apk add --no-cache openssl
-COPY bin/castai-cluster-controller /usr/local/bin/castai-cluster-controller
+ARG TARGETARCH amd64
```
r0kas, line 3: what about arm Targetarch? Separate dockerfile is needed for that?

varnastadeus (approved):

Dockerfile
andrejatcastai: Interesting, I thought ARG usage is like this (as per https://docs.docker.com/engine/reference/builder/#default-values):
ARG var=default_value

Does it work without `=`?

@r0kas no need for dedicated Docker file. Here a default value is defined. @aldor007 please correct me if I am wrong, I suppose TARGETARCH gets 'injected' when Docker images are build.

aldor007: Yes, TARGETARCH is injected by docker during the build

example
```shell
$ docker buildx build --platform linux/amd64,linux/arm64  -f Dockerfile . -t 

[...]
=> [linux/arm64 2/2] COPY bin/spot-handler-arm64 /usr/local/bin/spot-handler                                                                                                                                                      0.3s
=> [linux/amd64 2/2] COPY bin/spot-handler-amd64 /usr/local/bin/spot-handler                                                                                                                                                      0.2s
=> exporting to image   
```

andrejatcastai (approved):

PR #52, opened 2022-11-24
Send Node Architecture with AKS Init Data

fals:
- Added new field to request
- Added unit test

### DIFF
--- a/castai/types.go
+++ b/castai/types.go
@@ -186,7 +186,15 @@ func (c *ChartSource) Validate() error {
 	return nil
 }
 
+type NodeArchitecture int32
+
+const (
+	NodeArchitectureAmd64 NodeArchitecture = 0
+	NodeArchitectureArm64 NodeArchitecture = 1
+)
+
 type AKSInitDataRequest struct {
-	CloudConfigBase64       string `json:"cloudConfigBase64"`
-	ProtectedSettingsBase64 string `json:"protectedSettingsBase64"`
+	CloudConfigBase64       string            `json:"cloudConfigBase64"`
+	ProtectedSettingsBase64 string            `json:"protectedSettingsBase64"`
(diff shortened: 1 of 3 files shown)


castai/types.go
```diff
@@ -186,7 +186,15 @@ func (c *ChartSource) Validate() error {
 	return nil
 }
 
+type NodeArchitecture int32
+
+const (
+	NodeArchitectureAmd64 NodeArchitecture = 0
+	NodeArchitectureArm64 NodeArchitecture = 1
```
varnastadeus, line 193: why not just sending `arm64/amd64` instead? 🤔

fals: I designed the api to not rely on strings, I think it is easier to transform and make operations.

varnastadeus: But it is much easier to reason about as 0/1 doesn't says much and you need to do additional mapping everywhere down the pipeline

fals: changed as suggested

castai/types.go
```diff
@@ -187,6 +187,7 @@ func (c *ChartSource) Validate() error {
 }
 
 type AKSInitDataRequest struct {
-	CloudConfigBase64       string `json:"cloudConfigBase64"`
-	ProtectedSettingsBase64 string `json:"protectedSettingsBase64"`
+	CloudConfigBase64       string  `json:"cloudConfigBase64"`
+	ProtectedSettingsBase64 string  `json:"protectedSettingsBase64"`
+	Architecture            *string `json:"architecture,omitempty"`
```
varnastadeus, line 192: why its nullable?

fals: IMO Everything new should be optional. What is the concern?

fals: Just added as suggested

varnastadeus (approved):

PR #58, opened 2023-02-03
Fixing package vulnerabilities go-restful and helm

fals:


.github/workflows/fossa.yaml
```diff
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - main
-
+  schedule:
+    - cron: "0 */12 * * *"
```
andrejatcastai, line 8: So FOSSA action will run BOTH on merge to master AND according to the schedule?

Dockerfile
```diff
@@ -1,5 +1,4 @@
-FROM alpine:3.17.1
-RUN apk add --no-cache openssl
+FROM gcr.io/distroless/static:nonroot
```
andrejatcastai, line 1: Could you create a ticket for migrating other components?

andrejatcastai (approved):

.github/workflows/fossa.yaml
fals: @andrejatcastai this schedule ensures that every 12 hours the FOSSA action will run on the main branch, so we will proactively get new things on FOSSA instead of waiting code changes.

aldor007 (approved):

PR #60, opened 2023-02-08
draining tunning

aldor007:
* increase delete tries and timeout
* change chunk size to evict/delete to  be based on the amount of pods
* take into account drain action creation time


actions/drain_node_handler.go
```diff
@@ -202,10 +237,20 @@ func (h *drainNodeHandler) listNodePodsToEvict(ctx context.Context, node *v1.Nod
...
 	}
 
-	podsToEvict := lo.Filter(pods.Items, func(pod v1.Pod, _ int) bool {
-		return !isDaemonSetPod(&pod) && !isStaticPod(&pod)
-	})
+	podsToEvict := make([]v1.Pod, 0)
+	castPods := make([]v1.Pod, 0)
+	// Evict CAST PODs as last ones
+	for _, p := range pods.Items {
+		if p.Namespace == h.cfg.castNamespace && !isDaemonSetPod(&p) && !isStaticPod(&p) {
```
fals, line 244: Wouldn't be also interesting filter the pods that are possible to evict?

aldor007: Yes, good idea. I would add filter for pod that are already in deleting state/phase
https://github.com/kubernetes/kubectl/blob/31af32762da0400245ada0c342327da4b78eb182/pkg/drain/filters.go#L247

fals: Yeah good one, also reducing the number of calls to api-server would improve the throughput of controller. We should also remove any other status we can't evict, like Unknown?

WDYT?

aldor007: Also, good idea, will validate it tomorrow.

aldor007: I won't use `PodUnknown` as 
`// Deprecated: It isn't being set since 2015 (74da3b14b0c0f658b3bb8d2def5094686d0e9095)
`

actions/drain_node_handler.go
```diff
@@ -39,10 +44,29 @@ func newDrainNodeHandler(log logrus.FieldLogger, clientset kubernetes.Interface)
...
+func (h *drainNodeHandler) getDrainTimeout(action *castai.ClusterAction) time.Duration {
+	timeSinceCreated := time.Since(action.CreatedAt)
+	drainTimeout := time.Duration(action.ActionDrainNode.DrainTimeoutSeconds) * time.Second
+	// Allow to have 2 minutes grace period for newer actions.
+	if timeSinceCreated < 2*time.Minute {
+		return drainTimeout
+	}
+
+	// Remove time required for polling the action.
+	timeSinceCreated = timeSinceCreated - 100*time.Second
```
r0kas, line 65: 100 Seconds is quite a lot?
Also we have actions priority which might impact when action is picked up during times when there's a long queue of actions.

actions/drain_node_handler.go
```diff
@@ -39,10 +44,29 @@ func newDrainNodeHandler(log logrus.FieldLogger, clientset kubernetes.Interface)
...
 }
 
+// getDrainTimeout returns drain timeout adjusted to action creation time.
+// After 2 minutes grace period action timeout is adjusted using the time passed between action creation and now
+// the result is clamped between 60s and the requested timeout.
+func (h *drainNodeHandler) getDrainTimeout(action *castai.ClusterAction) time.Duration {
+	timeSinceCreated := time.Since(action.CreatedAt)
+	drainTimeout := time.Duration(action.ActionDrainNode.DrainTimeoutSeconds) * time.Second
+	// Allow to have 2 minutes grace period for newer actions.
+	if timeSinceCreated < 2*time.Minute {
```
andrejatcastai, line 60: There are 3 constants used:
2*time.Minute
100*time.Second (roundtrip time)
60*time.Second

Would be easier to follow if dedicated const are defined with descriptive names ;)

actions/drain_node_handler.go
```diff
@@ -39,10 +44,29 @@ func newDrainNodeHandler(log logrus.FieldLogger, clientset kubernetes.Interface)
...
+		return drainTimeout
+	}
+
+	// Remove time required for polling the action.
+	timeSinceCreated = timeSinceCreated - 100*time.Second
+
+	return lo.Clamp(drainTimeout-timeSinceCreated, 60*time.Second, time.Duration(action.ActionDrainNode.DrainTimeoutSeconds)*time.Second)
```
r0kas, line 67: Could you explain how this works? Clamp doesn't ring a bell :)

actions/drain_node_handler.go
```diff
@@ -139,8 +163,18 @@ func (h *drainNodeHandler) evictNodePods(ctx context.Context, log logrus.FieldLo
 	}
 
 	log.Infof("evicting %d pods", len(pods))
+	if len(pods) == 0 {
+		return nil
+	}
+	groupVersion, err := drain.CheckEvictionSupport(h.clientset)
```
r0kas, line 169: If eviction is not supported, but force drain is selected, should we care?

fals (approved):

actions/drain_node_handler.go
```diff
@@ -202,10 +239,26 @@ func (h *drainNodeHandler) listNodePodsToEvict(ctx context.Context, node *v1.Nod
...
+		if !p.ObjectMeta.DeletionTimestamp.IsZero() &&
+			int(time.Since(p.ObjectMeta.GetDeletionTimestamp().Time).Seconds()) > h.cfg.skipDeletedTimeoutSeconds {
+			continue
+		}
+
+		if p.Namespace == h.cfg.castNamespace && !isDaemonSetPod(&p) && !isStaticPod(&p) {
+			castPods = append(castPods, p)
+		}
 
+		if !isDaemonSetPod(&p) && !isStaticPod(&p) {
```
r0kas, line 256: I've read somewhere that Completed pods fail eviction. We should just skip those if they are part of the list.

actions/drain_node_handler.go
aldor007: This is not about checking if eviction is supported but with  version is supported. 
> Note: policy/v1 Eviction is available in v1.22+. Use policy/v1beta1 with prior releases.

I don't think that we have customer with K8s cluster that doesn't support eviciton.

actions/drain_node_handler.go
aldor007: I was checking this by scheduling jobs in my cluster and eviction was working. But I can add the rule for it.

actions/drain_node_handler.go
aldor007: Clamp will return a number in the given range. 

```python
function math.clamp(number, min, max)
	if number < min then
		return min
	elseif number > max then
		return max    
	end
	return number
end
```

https://wiki.multitheftauto.com/wiki/Math.clamp

actions/drain_node_handler.go
aldor007: Will do

actions/drain_node_handler.go
```diff
@@ -12,24 +12,35 @@ import (
...
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubectl/pkg/drain"
 
 	"github.com/castai/cluster-controller/castai"
 )
 
+const (
+	minDrainTimeout   = 60               //
+	actionGracePeriod = 2 * time.Minute  // Arbitrary chosen timeout after which action creation time will be taken into account when calculating drain timeeout
+	roundtirpTime     = 10 * time.Second //
```
fals, line 30: typo, roundtripTime

actions/drain_node_handler.go
r0kas: I believe we shouldn't have this `if` statement. This was the reason that we were cycling the same 2 minutes timeout all the time when controller was being restarted every 30 seconds.
Actual timeout should be a function:
`actualTimeout := drainTimeout - timeSinceCreated`
This could be rounded to 10 seconds intervals to account for roundtrips.

The only caveat here is if actions priority doesn't pick up this action in time due to action priorities.

actions/drain_node_handler.go
```diff
@@ -39,10 +50,29 @@ func newDrainNodeHandler(log logrus.FieldLogger, clientset kubernetes.Interface)
...
 }
 
+// getDrainTimeout returns drain timeout adjusted to action creation time.
+// After 2 minutes grace period action timeout is adjusted using the time passed between action creation and now
+// the result is clamped between 60s and the requested timeout.
+func (h *drainNodeHandler) getDrainTimeout(action *castai.ClusterAction) time.Duration {
+	timeSinceCreated := time.Since(action.CreatedAt)
+	drainTimeout := time.Duration(action.ActionDrainNode.DrainTimeoutSeconds) * time.Second
+	// Allow to have 2 minutes grace period for newer actions.
+	if timeSinceCreated < actionGracePeriod {
```
r0kas, line 66: I believe we shouldn't have this if statement. This was the reason that we were cycling the same 2 minutes timeout all the time when controller was being restarted every 30 seconds.
Actual timeout should be a function:
actualTimeout := drainTimeout - timeSinceCreated
This could be rounded to 10 seconds intervals to account for roundtrips.

The only caveat here is if actions priority doesn't pick up this action in time due to action priorities.

actions/drain_node_handler.go
```diff
@@ -12,24 +12,35 @@ import (
...
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubectl/pkg/drain"
 
 	"github.com/castai/cluster-controller/castai"
 )
 
+const (
+	minDrainTimeout   = 60               // Minimal pod drain timeout
```
r0kas, line 28: This might be too big if controller is being restarted every 30seconds. This could be much shorter if we'd know that we already sent evict some time ago. As it's async operation.

actions/drain_node_handler.go
r0kas: No need to send evict calls for Completed pods. This will lower the API calls count in some scenarios.

aldor007: Added

r0kas (approved):
LGTM

actions/drain_node_handler_test.go
```diff
@@ -268,6 +341,7 @@ func prependEvictionReaction(t *testing.T, c *fake.Clientset, success bool) {
 
 			go func() {
 				err := c.CoreV1().Pods(eviction.Namespace).Delete(context.Background(), eviction.Name, metav1.DeleteOptions{})
+				fmt.Println("EEEERRR", err)
```
aldor007, line 344: Fixme!

andrejatcastai (approved):

PR #67, opened 2023-04-07
Log when evicting CAST AI components

fals:


Narunas-K (approved):

actions/drain_node_handler.go
```diff
@@ -352,6 +354,20 @@ func (h *drainNodeHandler) deletePod(ctx context.Context, options metav1.DeleteO
...
+		return
+	}
+
+	castPodsNames := make([]string, 0, len(castPods))
+	for _, p := range castPods {
+		castPodsNames = append(castPodsNames, p.Name)
+	}
+	joinedPodNames := strings.Join(castPodsNames, ", ")
+
+	logrus.Warnf("evicting CAST AI pods: %s", joinedPodNames)
```
aldor007, line 368: I think we will not have all fields from here, right?
https://github.com/castai/cluster-controller/blob/main/main.go#L129-L134

fals: we already have something else in the handler, lots of fields
https://github.com/castai/cluster-controller/blob/main/actions/drain_node_handler.go#L82

fals: The main idea is just know how often we are killing castware and cross logs with other ones to debug. We have the node_id and other information from the handler.

actions/drain_node_handler_test.go
```diff
@@ -331,6 +331,24 @@ func TestGetDrainTimeout(t *testing.T) {
 	})
 }
 
+func TestLogCastPodsToEvict(t *testing.T) {
```
andrejatcastai, line 334: The code looks good to me. It's just tests look a bit meaningless here: there is no asserting that we logged what we had. More generally it is not even subject for our testing to assert the logs, so it is up to you if you want to keep this test ;)

fals: @andrejatcastai the test is jus making sure this method won't panic, I tried to include test cases for castware pods in the drain tests, it was flaky, tests failed randomly after adding a new pod to validate this single log, I decided to add a test, the assertion is basically passing.

actions/drain_node_handler.go
aldor007: Will this log have all fields from main.go or not?  If not it will not have a context of the controller.

actions/drain_node_handler_test.go
fals: @andrejatcastai added tests with assertion

aldor007 (approved):

PR #68, opened 2023-05-02
chore: remove aks handler

fals:
- remove aks handler
- vulnerabilities fixed with package update
- use distroless image


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.17.1
+FROM static-debian11
```
mjudeikis, line 1: Description says distroless but this is not?

fals: @mjudeikis what am I missing here? this list says it is 🤔 https://github.com/GoogleContainerTools/distroless

fals: We use this same image in other repos BTW

Narunas-K (approved):

Dockerfile
mjudeikis: You added a full path ;) problem with short path is that it defaults to docker hub and image might not be from the original publisher :)

PR #72, opened 2023-06-06
Add support for cordoning node as a part of PatchNode action

andrejatcastai:


actions/patch_node_handler.go
```diff
@@ -64,12 +65,17 @@ func (h *patchNodeHandler) Handle(ctx context.Context, action *castai.ClusterAct
 		return err
 	}
 
-	log.Infof("patching node, labels=%v, taints=%v, annotations=%v", req.Labels, req.Taints, req.Annotations)
+	unschedulable := "<nil>"
```
aldor007, line 68: This looks strange. %v can handle nil

andrejatcastai: %v handles nil when unschedulable is not provided. When it is provided then %v prints pointer value. It does not dereference when value is provided.

PR #73, opened 2023-07-12
CORE-3242: skip node delete on id mismatch

mjudeikis:
Skip node delete when node_id mismatch for case where cloud reuses the names

### DIFF
--- a/actions/delete_node_handler.go
+++ b/actions/delete_node_handler.go
@@ -60,25 +60,25 @@ func (h *deleteNodeHandler) Handle(ctx context.Context, action *castai.ClusterAc
 	})
 	log.Info("deleting kubernetes node")
 
-	current, err := h.clientset.CoreV1().Nodes().Get(ctx, req.NodeName, metav1.GetOptions{})
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info("node not found, skipping delete")
-			return nil
-		}
-		return fmt.Errorf("error getting node %w", err)
-	}
-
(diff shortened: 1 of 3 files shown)


castai/types.go
```diff
@@ -8,6 +8,10 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+const (
+	LabelProvisionerCastAINodeID = "provisioner.cast.ai/node-id:"
```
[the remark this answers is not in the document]
mjudeikis: yes! good catch! And test was passing as I was reusing same value in test :D :D

aldor007 (approved):

actions/delete_node_handler.go
```diff
@@ -60,9 +60,25 @@ func (h *deleteNodeHandler) Handle(ctx context.Context, action *castai.ClusterAc
 	})
 	log.Info("deleting kubernetes node")
 
+	current, err := h.clientset.CoreV1().Nodes().Get(ctx, req.NodeName, metav1.GetOptions{})
```
aldor007, line 63: Maybe let's add retry for the get? Like for delete call?

actions/delete_node_handler.go
```diff
@@ -60,9 +60,25 @@ func (h *deleteNodeHandler) Handle(ctx context.Context, action *castai.ClusterAc
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("node not found, skipping delete")
+			return nil
+		}
+		return fmt.Errorf("error getting node %w", err)
```
varnastadeus, line 69:
```suggestion
		return fmt.Errorf("getting node %w", err)
```

PR #75, opened 2023-07-20
feat: tune leader elect timeouts

aldor007:


main.go
```diff
@@ -217,7 +219,7 @@ func runWithLeaderElection(
 			},
 			OnStoppedLeading: func() {
 				log.Infof("leader lost: %s", id)
-				os.Exit(0)
+				logrus.Exit(1)
```
fals, line 222: Any reason why we would return a non-zero exit signal here?

Narunas-K (approved):
LGTM

PR #76, opened 2023-08-02
feat: add the ability to patch deployments to change pod requests and replicas

UndeadRat22:


actions/kubernetes_helpers.go
```diff
@@ -15,31 +16,45 @@ import (
...
+			_, err := clientset.CoreV1().Nodes().Patch(ctx, node.Name, patchType, data, options)
+			return err
+		},
+	)
+}
+
+func patchObject[T any](
+	ctx context.Context,
+	object *T,
+	modify func(*T) error,
```
varnastadeus, line 31: do we need `modify` as part of generic method or can we pass already modified object straight to patch?

UndeadRat22: would need to deep copy the object, the implemation is 
1. marshal current 
2. modify current
3. marshal modified
4. get diff

PR #77, opened 2023-08-04
feat: dynamic patch action

UndeadRat22:
This action allows any resource that the cluster-controller has permission for.

### DIFF
--- a/actions/errors.go
+++ b/actions/errors.go
@@ -1,7 +1,0 @@
-package actions
-
-import "fmt"
-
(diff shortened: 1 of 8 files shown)


actions/errors.go
```diff
@@ -0,0 +1,7 @@
+package actions
+
+import "fmt"
+
+func newUnexpectedTypeErr(value interface{}, expectedType interface{}) error {
```
varnastadeus, line 5: nitpick, just move this under actions.go? 🙏

varnastadeus (approved):

PR #85, opened 2023-09-13
feat: add resource create/delete actions

varnastadeus:


actions/create_handler.go
```diff
@@ -0,0 +1,118 @@
...
+	if newObj.GetNamespace() == "" {
+		return errors.New("object namespace is missing")
+	}
+
+	log := h.log.WithFields(logrus.Fields{
+		"id":       action.ID,
+		"action":   reflect.TypeOf(action.Data()).String(),
+		"group":    req.Group,
+		"resource": req.Resource,
+		"version":  req.Version,
```
mjudeikis, line 53: I would construct proper GVR resource as single field. Easier to read in logs...

mjudeikis (approved):
I would just tune logging as GRV is not quite intended to be separate fields in real world... everybody is used to apps/v1 or `stuff.core/v1alpha1`

actions/create_handler.go
varnastadeus: makes sense, had similar idea initially but kept it consistent with `patch` handler. Will change it there as well 🙏

actions/create_handler.go
```diff
@@ -0,0 +1,118 @@
...
+			return err
+		}
+
+		patch, err := jsonpatch.CreateMergePatch(original, modified)
+		if err != nil {
+			return fmt.Errorf("creating patch: %w", err)
+		}
+
+		// If resources are identical, patch will be equal '{}'.
+		if len(patch) <= 2 {
```
UndeadRat22, line 103: if it will be equal to {}, we can just say if patch == '{}' return nil

UndeadRat22 (approved):

PR #94, opened 2024-02-08
CSU-1004: handle http timeouts

ValyaB:


stgleb (approved):
LGTM

main.go
```diff
@@ -65,13 +68,20 @@ func main() {
 		},
 	}
 
+	log := logrus.WithFields(logrus.Fields{})
+
+	// Create a new client to communicate with CAST AI API.
+	deltaHTTPClient, err := castai.NewDefaultClient(cfg.API.URL, cfg.API.Key, logger.Level, binVersion, defaultPollTimeout)
```
andrejatcastai, line 74: I suppose this is copy-paste typo

main.go
```diff
@@ -65,13 +68,20 @@ func main() {
...
+	if err != nil {
+		log.Fatalf("new http client failed: %v", err)
+	}
+
 	client := castai.NewClient(
 		logger,
-		castai.NewDefaultClient(cfg.API.URL, cfg.API.Key, logger.Level, binVersion),
+		deltaHTTPClient,
```
andrejatcastai, line 81: I can see it is not your change, but why here it is `logger`, whereas above it `log := logrus.WithFields(logrus.Fields{})`?

castai/client.go
```diff
@@ -33,17 +36,54 @@ func NewClient(log *logrus.Logger, rest *resty.Client, clusterID string) Client
...
+			Timeout: 15 * time.Second,
+		}).DialContext,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   5 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
+
+	t2, err := http2.ConfigureTransports(t1)
```
andrejatcastai, line 73: We are not using http2 in controller. Could you elaborate why http2?

ValyaB: we are, it is default from resty library

main.go
ValyaB: Where? Can you explain, please

ValyaB:
Without configured tls, http2 is not used,  and 5m timeout is by server design

main.go
apasyniuk: `deltaHTTPClient` we don't have deltas in the controller

main.go
andrejatcastai: There are two logger objects created (logger and log). Could you explain why it is needed?

castai/client.go
andrejatcastai: I could not find Resty is calling `http2.ConfigureTransports`. Moreover to my understanding Resty relies on `http` (even though it sets `ForceAttemptHTTP2` to true) package: https://github.com/go-resty/resty/blob/v2/go.mod#L6

main.go
ValyaB: It probably was added to avoid additional fields and log levels from env when printing fatal in the main.

PR #100, opened 2024-02-26
KUBE-194: permissions might not be upgraded yet so silence forbidden error

furkhat:
Defensive programming. If `casai-cluster-controller` was upgraded but permission to Patch `node/status` were not added we don't won't to fail the whole action.


stgleb (approved):
LGTM

PR #103, opened 2024-02-29
Give enough time to sync clientset's cache on new Node 3.5 sec -> 10 sec

furkhat:
Change retries timeout from ~3.5sec to 10sec.

### DIFF
--- a/actions/patch_node_handler.go
+++ b/actions/patch_node_handler.go
@@ -56,10 +56,6 @@ func (h *patchNodeHandler) Handle(ctx context.Context, action *castai.ClusterAct
 		"id":        action.ID,
 	})
 
-	// on GKE we noticed that sometimes the node is not found, even though it is in the cluster
-	// as result was returned from watch. But subsequent get request returns not found.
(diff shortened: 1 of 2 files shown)


actions/patch_node_handler.go
```diff
@@ -56,10 +56,6 @@ func (h *patchNodeHandler) Handle(ctx context.Context, action *castai.ClusterAct
 		"id":        action.ID,
 	})
 
-	// on GKE we noticed that sometimes the node is not found, even though it is in the cluster
```
stgleb, line 59: Why you meved that commend to function body?

apasyniuk (approved):

actions/patch_node_handler.go
furkhat: The comment I moved mentions retries but adjacent lines do not have retries because retry happens inside of  "get node" function. 
Would you agree that keeping the comment close the code it relates would be better?

That said I wouldn't have necessarily moved it but since I  put in the comment the real cause why client doesn't find new node right away (i.e. delayed cache sync) I decided to move it where it suits more.

PR #111, opened 2024-04-12
Ideomatic Go interface for log exporter

furkhat:


castai/client.go
```diff
@@ -23,11 +24,19 @@ const (
...
+type client struct {
+	log       *logrus.Logger
+	rest      *resty.Client
+	clusterID string
+}
+
+// client must satisfy logexporter.Sender.
+var _ logexporter.Sender = new(client)
+
+func NewClient(log *logrus.Logger, rest *resty.Client, clusterID string) *client {
```
ValyaB, line 39: Public func returns a public struct
make client public

stgleb: type itself yes, but fields are still private, aren't they?

logexporter/exporter.go
```diff
@@ -20,20 +18,34 @@ type Exporter interface {
 	Wait()
 }
 
-func NewExporter(logger *logrus.Logger, client castai.Client) Exporter {
-	return &exporter{
-		logger: logger,
-		client: client,
-		wg:     sync.WaitGroup{},
-	}
```
stgleb, line 28: Maybe LogSender than?

stgleb (approved):
LGTM

castai/client.go
furkhat: Yep I will make the type public and fields will stay private.

logexporter/exporter.go
furkhat: I think linter was complaining that there is a tautology. It was saying that `logexporter.Sender` is better than `logexporter.LogSender`.
I will double check and change if it doesn't complain.

PR #113, opened 2024-04-18
Kube-333: Unify the retry behavior around k8s.io/apimachinery/pkg/util/wait

Tsonov:
... so we can remove the dependency on another backoff library. 

A "functional" change is that we also do more logging now on retries. The rest should match the current behavior in the code. 

To do so, a small package is added. It has similar behavior to the backoff library but based on `wait` instead. 

Unfortunately the wait package has a few quirks so I have tried to make the helpers easy to use and obvious while 

Notably:

1. There is no "exponential backoff increasing forever" like in the backoff package, there is always a max number of increases. The forever functions in `wait` don't use a backoff at all or manually reset backoffs. I think int.MaxValue is a simple alternative as realistically we would never wait for 2,000,000 exponential increases...
2. The wait.Backoff.Steps value means two things in the library - _how many times to increase an exponential backoff_ and _how many times to execute_. 

The second point is a confusing design choice IMO. So I see two ways to design our helpers:

Our backoff constructors return "forever" backoffs but then either Context or WithMaxRetries should be used to avoid infinite loops.

-or-

Our backoff constructors do not set the number of steps but then callers must remember to set the Retry count every time or it will run only once (which is different behavior from `backoff` library that is used in other repos). 

I have chosen the first one as it feels more consistent and I think the risk for mistake is lower but open for opinion. 

Overall, it looks like the two libraries had different philosophies behind them :) 

The new helper is under `waitext` as extensions to the wait package. 

Tests ran:

- [ x] Onboard cluster to phase 2 and change image
- [ x] Do rebalance that deleted 3 nodes and added 3 cast.ai nodes
- [x] Delete a node from UI
- [ x] Force autoscaler to add node by scaling workloads on cluster
- [x] Deleted too nodes from UI to force under-provisioning and adding nodes


actions/approve_csr_handler.go
```diff
@@ -122,25 +125,23 @@ func (h *approveCSRHandler) getInitialNodeCSR(ctx context.Context, log logrus.Fi
 	var cert *csr.Certificate
 	var err error
 
-	logRetry := func(err error, _ time.Duration) {
-		log.Warnf("getting initial csr, will retry: %v", err)
-	}
-	b := backoff.WithMaxRetries(backoff.NewExponentialBackOff(), 3)
-	err = backoff.RetryNotify(func() error {
+	// TODO: Shouldn't we pass ctx here, too?
```
Tsonov, line 128: I left this here to discuss - We use a context in the function above. Shouldn't we pass the context here as well? 
Original code did not do that.

actions/kubernetes_helpers.go
```diff
@@ -65,27 +72,29 @@ func getNodeForPatching(ctx context.Context, log logrus.FieldLogger, clientset k
...
-		log.Warnf("getting node, will retry: %v", err)
-	}
+	// So we give enough time for cache to sync - ~10s max.
+
 	var node *v1.Node
-	b := backoff.NewExponentialBackOff()
-	b.MaxElapsedTime = 10 * time.Second
-	err := backoff.RetryNotify(func() error {
+
+	boff := waitext.WithMaxRetries(waitext.DefaultExponentialBackoff(), 5)
```
Tsonov, line 79: I changed this because `wait` does not have a way to say "Stop after X duration" directly. It uses contexts for that but then here we would surface the context.Cancelled error and not the `Nodes().Get()` error, which breaks calling code.

I think this is an OK compromise, it still waits for ~10s tops and surfaces the inner error (unless the overall context is cancelled of course).

actions/actions.go
```diff
@@ -132,16 +132,19 @@ func (s *service) doWork(ctx context.Context) error {
 		iteration int
 	)
 
-	b := backoff.WithContext(backoff.WithMaxRetries(backoff.NewConstantBackOff(5*time.Second), 3), ctx)
-	errR := backoff.Retry(func() error {
-		iteration++
```
ValyaB, line 137: if we are removing this, why we are printing it then?

actions/actions.go
```diff
@@ -132,16 +132,19 @@ func (s *service) doWork(ctx context.Context) error {
...
+	errR := waitext.RetryWithContext(ctx, boff, func(ctx context.Context) error {
 		actions, err = s.castAIClient.GetActions(ctx, s.k8sVersion)
 		if err != nil {
 			s.log.Errorf("polling actions: get action request failed: iteration: %v %v", iteration, err)
 			return err
 		}
 		return nil
-	}, b)
+	}, func(err error) {
+		s.log.Warnf("polling actions failed, will retry: %v", err)
```
ValyaB, line 145: do we need to print it again (The first time, it was in the 140th line)

actions/actions.go
Tsonov: Good point, I didn't notice we already have "log-per-retry" here so we double-log now. Will fix.

internal/waitext/doc.go
```diff
@@ -0,0 +1,6 @@
+package waitext
```
furkhat, line 1: Hey I think `internal` dir is not needed. 
Internal folder helps forbidding usage of packages that are meant to be used within the local codebase. I think it is most useful for protecting clients from unwanted API dependency.
Since cluster-controller is never a dependency (nobody will ever do `go mod add github.com/castai/cluster-controller`) I think we don't need `internal` folder here. What do you think?

actions/approve_csr_handler_test.go
```diff
@@ -4,11 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/google/uuid"
 	"sync/atomic"
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
+
```
furkhat, line 12: We tried to group imports into 3 groups.
(
builtin

external

internal
)

Could you please join this with external imports?

internal/waitext/error.go
```diff
@@ -0,0 +1,22 @@
+package waitext
+
+// NonTransientError is used in conjuction with Retry or RetryWithContext
+// to allow the operation to signal that retrying is pointless
+// It works the same as PermanentError in github.com/cenkalti/backoff
+type NonTransientError struct {
```
furkhat, line 6: Please consider minimizing the amount of abstraction we introduce. We could just make retry function return `(error, bool)`. The 2nd bool indicates whether the function should be retried or not.

internal/waitext/extensions.go
```diff
@@ -0,0 +1,153 @@
...
+//   - the context is cancelled
+//
+// The end result is the final error observed when calling operation() or nil if successful or context.Err() if the context was cancelled.
+// If retryNotify is passed, it is called when making retries.
+// Caveat: this function is similar to wait.ExponentialBackoff but has some important behavior differences like at-least-one execution and retryable errors
+func RetryWithContext(ctx context.Context, backoff wait.Backoff, operation func(context.Context) error, retryNotify func(error)) error {
+	return retryCore(ctx, backoff, operation, retryNotify)
+}
+
+func retryCore(ctx context.Context, backoff wait.Backoff, operation func(context.Context) error, retryNotify func(error)) error {
```
furkhat, line 112: Hey I can't stop but feel this is too complex. I think we can just have one retry functions and that would satisfy all use cases. Could you consider this?

furkhat (changes requested):
Let's try to minimise the amount of code/abstraction we introduce.

actions/actions.go
Tsonov: I think I implemented your goal, we now log an error on every failed attempt and it has the iteration and actual error observed - please check.

internal/waitext/doc.go
Tsonov: Sure, do you think we should move it under root (`cluster-controller/waitext/*`) or use `pkg` (`cluster-controller/pkg/waittext/*`). Second is probably more idiomatic but we don't do it for other packages in this repo. 

No preference from my side TBH

furkhat: I think under root because other packages are like that.

actions/approve_csr_handler.go
furkhat: Thanks that's a good catch. I think it is a good idea if we pass it.

actions/actions.go
```diff
@@ -132,16 +132,19 @@ func (s *service) doWork(ctx context.Context) error {
...
 		actions, err = s.castAIClient.GetActions(ctx, s.k8sVersion)
 		if err != nil {
-			s.log.Errorf("polling actions: get action request failed: iteration: %v %v", iteration, err)
-			return err
+			return true, err
 		}
-		return nil
-	}, b)
+		return false, nil
+	}, func(err error) {
```
furkhat, line 144: Could we make the logging of the error the default? Then we don't need to pass 2nd "logNotify" function. If someone wants a custom log message they could just wrap returning error with as much details as they need using `fmt.Errorf`. What do you think?

helm/chart_loader.go
```diff
@@ -86,8 +102,8 @@ func (cl *remoteChartLoader) fetchArchive(ctx context.Context, archiveURL string
 	return archiveResp, nil
 }
 
-func defaultBackoff(ctx context.Context) backoff.BackOffContext {
-	return backoff.WithContext(backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), 5), ctx)
+func defaultBackoff() wait.Backoff {
+	return waitext.NewConstantBackoff(1 * time.Second)
```
furkhat, line 106: I feel like this is not very useful function as well. Could you consider using `waitext` directly here as well?

waitext/extensions.go
```diff
@@ -0,0 +1,118 @@
...
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	DefaultInitialInterval     = 500 * time.Millisecond
+	DefaultRandomizationFactor = 0.5
+	DefaultMultiplier          = 1.5
+	DefaultMaxInterval         = 60 * time.Second
```
furkhat, line 15: Are this constants used outside of the package? Do you think we could not export them?

waitext/extensions.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Forever should be used to simulate infinite retries or backoff increase.
+	// Usually it's wise to have a context with timeout to avoid an infinite loop.
+	Forever = math.MaxInt32
+)
+
+// NewExponentialBackoff creates a backoff that increases the delay between each step based on a factor.
+// If maxInterval is positive, then the wait duration will not exceed its value while increasing.
+// Essentially at step N the wait is min(initialInterval*factor^(N-1), maxInterval)
+func NewExponentialBackoff(initialInterval time.Duration, factor float64, maxInterval time.Duration) wait.Backoff {
```
furkhat, line 25: It seems like this function is not used outside the package. Would it be okay to delete it?

waitext/extensions.go
```diff
@@ -0,0 +1,118 @@
...
+// The operation will not be retried anymore if
+//
+//   - retries reaches 0
+//
+//   - the context is cancelled
+//
+// The end result is the final error observed when calling operation() or nil if successful or context.Err() if the context was cancelled.
+// If retryNotify is passed, it is called when making retries.
+// Caveat: this function is similar to wait.ExponentialBackoff but has some important behavior differences like at-least-one execution and retryable errors
+func RetryWithContext(ctx context.Context, backoff wait.Backoff, retries int, operation func(context.Context) (bool, error), retryNotify func(error)) error {
```
furkhat, line 80: I think backoff called the function `RetryWithContext` because there is a function called `Retry` (which they deprecated I think). Since we have just one function could we simply call it `Retry`?

waitext/extensions.go
```diff
@@ -0,0 +1,95 @@
...
+	for retries > 0 {
+		// Notify about expected retry
+		if retryNotify != nil {
+			retryNotify(lastErr)
+		}
+
+		waitInterval := backoff.Step()
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
furkhat, line 81: Do you remember that Valentyna mentioned that it would be nice to see the last error if context is cancelled due to pod termination? This helps investigating liveness probe issues. Would this ensure that the last error will return instead?
```suggestion
			break
```

actions/approve_csr_handler.go
```diff
@@ -122,25 +131,28 @@ func (h *approveCSRHandler) getInitialNodeCSR(ctx context.Context, log logrus.Fi
...
 }
 
-func newApproveCSRExponentialBackoff() *backoff.ExponentialBackOff {
-	b := backoff.NewExponentialBackOff()
-	b.Multiplier = 2
-	b.MaxElapsedTime = 4 * time.Minute
-	b.Reset()
+func newApproveCSRExponentialBackoff() wait.Backoff {
+	b := waitext.DefaultExponentialBackoff()
+	b.Factor = 2
```
furkhat, line 156: Not to run this forever and preserve old `b.MaxElapsedTime` behaviour.
```suggestion
	b.Factor = 2
	b..Cap = 4 * time.Minute
```

waitext/extensions.go
```diff
@@ -0,0 +1,95 @@
...
+// The operation will not be retried anymore if
+//
+//   - retries reaches 0
+//
+//   - the context is cancelled
+//
+// The end result is the final error observed when calling operation() or nil if successful or context.Err() if the context was cancelled.
+// If retryNotify is passed, it is called when making retries.
+// Caveat: this function is similar to wait.ExponentialBackoff but has some important behavior differences like at-least-one execution and retryable errors
+func Retry(ctx context.Context, backoff wait.Backoff, retries int, operation func(context.Context) (bool, error), retryNotify func(error)) error {
```
furkhat, line 61: It looks that wait.Backoff has information on how many or for how long to retry. Could you explain your motivation behond `retries` argument? Specifically why `backoff.Steps` is not enough?

actions/kubernetes_helpers.go
furkhat: > does not have a way to say "Stop after X duration" directly.

I think `Backoff.Cap` is used for this.

Tsonov: See comment in this thread https://github.com/castai/cluster-controller/pull/113#discussion_r1577728079 

TL;DR Backoff.Cap works on a single wait interval, not the overall operation - context.Timeout should be used for that.

furkhat (approved):

PR #115, opened 2024-05-21
KUBE-361: add CSR auto-approval

ValyaB:


csr/csr.go
```diff
@@ -268,6 +272,101 @@ func getNodeCSRV1Beta1(ctx context.Context, client kubernetes.Interface, nodeNam
...
+			sendCertificate(ctx, c, &Certificate{V1: csr})
+		}
+	}
+}
+func WatchAndApproveNodeCSRV1Beta1(ctx context.Context, log logrus.FieldLogger, client kubernetes.Interface, c chan *Certificate) error {
+	fs := fields.SelectorFromSet(fields.Set{
+		"spec.signerName": certv1beta1.KubeAPIServerClientKubeletSignerName,
+	})
+	ls := labels.SelectorFromSet(map[string]string{
+		castai.LabelManagedBy: castai.LabelValueManagedByCASTAI,
```
aldor007, line 320: I didn't see this selector on CSR.  Are you sure it is present?

ValyaB: yeah, good point, I will remove it

csr/csr.go
```diff
@@ -252,32 +247,116 @@ func getNodeCSRV1Beta1(ctx context.Context, client kubernetes.Interface, nodeNam
...
 	return nil, ErrNodeCertificateNotFound
 }
 
-func isNodeCSR(csrName string, csrRequest []byte, nodeName string) (bool, error) {
+func WatchAndApproveNodeCSR(ctx context.Context, log logrus.FieldLogger, client kubernetes.Interface, c chan *Certificate) error {
+	var w watch.Interface
+	var err error
+	w, err = client.CertificatesV1().CertificateSigningRequests().Watch(ctx, getOptions(certv1.KubeAPIServerClientKubeletSignerName))
+	if err != nil {
+		w, err = client.CertificatesV1beta1().CertificateSigningRequests().Watch(ctx, getOptions(certv1beta1.KubeAPIServerClientKubeletSignerName))
```
cirisked, line 270: will those watchers restarts in case of any network error?

csr/csr.go
```diff
@@ -252,32 +248,134 @@ func getNodeCSRV1Beta1(ctx context.Context, client kubernetes.Interface, nodeNam
...
+			if err != nil {
+				w, err = client.CertificatesV1beta1().CertificateSigningRequests().Watch(ctx, getOptions(certv1beta1.KubeAPIServerClientKubeletSignerName))
+				if err != nil {
+					return true, fmt.Errorf("fail to open v1 and v1beta watching client: %w", err)
+				}
+			}
+			return false, nil
+		},
+		func(err error) {
+			log.Warnf("retrying: %v", err)
```
Tsonov, line 285: Context on what is being retried would be useful here (more details in the message)

actions/approve_csr_handler.go
```diff
@@ -48,6 +52,19 @@ func (h *approveCSRHandler) Handle(ctx context.Context, action *castai.ClusterAc
 		actionIDLogField: action.ID,
 	})
 
+	if req.AllowAutoApprove != nil && *req.AllowAutoApprove {
+		go h.RunAutoApprove(ctx)
```
andrejatcastai, line 56: Could you describe concurrency design in the comments? There are:
- goroutine
- mutex
- multiple g.Go(...)

I wonder if all this is necessary.

csr/csr.go
ValyaB: context is there, inside log fields
example
```
time="2024-05-21T19:21:39+02:00" level=warning msg="retrying: fail to open v1 and v1beta watching client: ups error" RunAutoApprove=auto-approve-csr
```

actions/approve_csr_handler.go
ValyaB: we have only one go routine with watching and approving messages
but multiple handlers can start it or stop it

ValyaB: and we have approved csr for the initial node join which should be approved not by Watcher

csr/csr.go
ValyaB: yes

actions/approve_csr_handler.go
```diff
@@ -34,6 +35,8 @@ type approveCSRHandler struct {
 	clientset              kubernetes.Interface
 	initialCSRFetchTimeout time.Duration
 	csrFetchInterval       time.Duration
+	cancelAutoApprove      context.CancelFunc
+	m                      sync.Mutex // protects cancelAutoApprove for concurrent access
```
andrejatcastai, line 39: May be rewording something like: `Used to make sure there is just one watcher running as it may be triggered from multiple CSR actions.`

cirisked (approved):

PR #118, opened 2024-05-30
add debug logs

ValyaB:


csr/csr.go
```diff
@@ -307,16 +308,35 @@ func WatchCastAINodeCSRs(ctx context.Context, log logrus.FieldLogger, client kub
...
-			if csrResult.Approved() || !isAutoApproveAllowedForNode(ctx, client, cn) {
+			if csrResult.Approved() {
+				continue
+			}
+			log.WithFields(logrus.Fields{
+				"csr":       name,
+				"node_name": cn,
+			}).Debugf("checking csr: %s, node: %s", name, cn)
+			enabled, err := isAutoApproveAllowedForNode(ctx, client, cn)
+			if !enabled || err != nil {
```
andrejatcastai, line 335: This error handling is not idiomatic. If err is returned then the rest of parameters are undefined. Here err is checked only if enabled is false.

andrejatcastai (approved):

PR #120, opened 2024-06-21
KUBE-424: Migrate from core/v1 events to events.k8s.io/v1

stgleb:


actions/create_event_handler.go
```diff
@@ -98,9 +110,29 @@ func (h *createEventHandler) Handle(ctx context.Context, action *castai.ClusterA
...
 
 	if _, err := h.clientset.CoreV1().Events(event.Namespace).Create(ctx, event, metav1.CreateOptions{}); err != nil {
-		return fmt.Errorf("creating event for ref %v: %w", req.ObjectRef, err)
+		return fmt.Errorf("creating core/v1/event for ref %v: %w", req.ObjectRef, err)
 	}
+	return nil
+}
 
+func (h *createEventHandler) handleEventV1(ctx context.Context, req *castai.ActionCreateEvent, namespace string) error {
+	event := &eventsv1.Event{
```
cirisked, line 119: should we also check for similar event here?

cirisked (changes requested):
Please, cover the failed attempt to create event in test also.

actions/create_event_handler.go
stgleb: nope, there are no search API where we can search for the same object and we can't list and filter for all events.

stgleb:
> Please, cover the failed attempt to create event in test also.

Added a test case, ven though we can't properly mock clientset to reply with error first.

actions/create_event_handler.go
cirisked: Cost are ok with it? Deduplication was done for a reason i believe.

stgleb: right now I am trying to incorporate event recorder that should address duplication issues.

actions/create_event_handler.go
```diff
@@ -2,139 +2,76 @@ package actions
...
+	log                logrus.FieldLogger
+	clientSet          kubernetes.Interface
+	recorderFactory    func(string) (record.EventBroadcaster, record.EventRecorder)
+	mu                 sync.RWMutex
+	eventNsBroadcaster map[string]record.EventBroadcaster
+	eventNsRecorder    map[string]record.EventRecorder
 }
 
 func (h *createEventHandler) Handle(ctx context.Context, action *castai.ClusterAction) error {
+	h.log.Infof("handle create event action: %s", action.Data())
```
aldor007, line 44: Do we want to log event body here?

actions/create_event_handler.go
```diff
@@ -2,139 +2,76 @@ package actions
...
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/kubernetes"
-
-	"github.com/castai/cluster-controller/castai"
+	typedv1core "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/record"
```
aldor007, line 13: From which k8s version is this package available?

stgleb: record package had been there for a while, EventRecorder interface is already 5 years old. New events API was promoted to GA in 1.19 https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/383-new-event-api-ga-graduation#implementation-history

actions/create_event_handler.go
stgleb: removed

aldor007 (approved):

cirisked (approved):

PR #125, opened 2024-06-28
KUBE-380: Fix clean up volume attachments

stgleb:


actions/delete_node_handler.go
```diff
@@ -156,10 +159,44 @@ func (h *deleteNodeHandler) Handle(ctx context.Context, action *castai.ClusterAc
 			if len(pods.Items) > 0 {
 				return true, fmt.Errorf("waiting for %d pods to be terminated on node %v", len(pods.Items), req.NodeName)
 			}
+			// Check if there are any volume attachments left for the node, but don't block on waiting for them.
+			volumeAttachments, err := h.clientset.StorageV1().VolumeAttachments().List(ctx, metav1.ListOptions{})
+			if !apierrors.IsForbidden(err) {
```
Tsonov, line 164: [Suggestion] This looks a bit easier to follow to me

```go
if err != nil && !apierrors.IsForbidden(err) {
    log.Warnf("unable to list volume attachments for node %q err: %v", req.NodeName, err)
} else if volumeAttachments != nil && len(...) > 0 {
 ...
}

```

btw why don't we also log Forbidden, do we have a common scenario for that? Since it'd be nice to debug that error as well

Tsonov (approved):

actions/delete_node_handler.go
```diff
@@ -156,10 +159,44 @@ func (h *deleteNodeHandler) Handle(ctx context.Context, action *castai.ClusterAc
 			if len(pods.Items) > 0 {
 				return true, fmt.Errorf("waiting for %d pods to be terminated on node %v", len(pods.Items), req.NodeName)
 			}
+			// Check if there are any volume attachments left for the node, but don't block on waiting for them.
+			volumeAttachments, err := h.clientset.StorageV1().VolumeAttachments().List(ctx, metav1.ListOptions{})
```
Tsonov, line 163: Does this filter by node? The filter is different from below and there we manually select the node. Wouldn't it get all volume attachments?

stgleb: yes, it would, unfortunately there are no node filters for node attachments.

actions/delete_node_handler.go
stgleb: done

PR #128, opened 2024-08-01
KUBE-479: Evict all pods even in case of PDB violations

Tsonov:
Main change - treat PDB violation errors during eviction as non-retryable. Sometimes the PDB can never be satisfied if we evict a pod so we'd be stuck in infinite retry. 

Additional changes:

- Fixed a bug where CAST pods were added to the eviction list twice - so they weren't evicted last correctly. 
- Changed the batching implementation with channel + workers. This does not guarantee 100% that pod operations can't get stuck but a single one won't block others (as it was before - a single infinite retry blocks all chunks after it). If we get #InfiniteRetries > #Tasks, we'd still be stuck but the chances are much lower. 
- Added some additional logging here and there for future debuggability

### DIFF
--- a/actions/drain_node_handler.go
+++ b/actions/drain_node_handler.go
@@ -80,6 +80,11 @@ func (h *drainNodeHandler) Handle(ctx context.Context, action *castai.ClusterAct
 		return fmt.Errorf("unexpected type %T for drain handler", action.Data())
 	}
 	drainTimeout := h.getDrainTimeout(action)
+	// CSU-1945 temporary reduction to force evicting even pods with violated PDBs
+	if req.Force {
+		drainTimeout = 5 * time.Second
+	}
+
 	log := h.log.WithFields(logrus.Fields{
 		"node_name":      req.NodeName,
 		"node_id":        req.NodeID,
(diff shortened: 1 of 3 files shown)


actions/drain_node_handler.go
```diff
@@ -209,19 +214,46 @@ func (h *drainNodeHandler) deleteNodePods(ctx context.Context, log logrus.FieldL
...
 	}
 
-	return nil
+	for range parallelTasks {
+		wg.Add(1)
+		go worker(taskChan)
+	}
+
+	for _, pod := range pods {
+		taskChan <- &pod
```
aldor007, line 250: Why using here a pointer?

actions/drain_node_handler.go
```diff
@@ -209,19 +214,46 @@ func (h *drainNodeHandler) deleteNodePods(ctx context.Context, log logrus.FieldL
...
+	)
+
+	h.log.Debugf("Starting %d parallel tasks for %d pods: [%v]", parallelTasks, len(pods), lo.Map(pods, func(t v1.Pod, i int) string {
+		return fmt.Sprintf("%s/%s", t.Namespace, t.Name)
+	}))
+
+	worker := func(taskChan <-chan *v1.Pod) {
+		for task := range taskChan {
+			if err := f(ctx, *task); err != nil {
+				taskErrsMx.Lock()
```
stgleb, line 236: maybe collect errors via error channel. There fore we can get rid-off both mutext and waitgroup. Just read all successes/errors.

actions/drain_node_handler.go
Tsonov: Fair, replaced with struct

actions/drain_node_handler.go
Tsonov: We'd still need waitgroup so we wait for all workers to finish (or we need to have a channel to signal when each one ends; I think waitgroup is simpler). Or maybe I am missing what you mean? 

For the errors, it can work either way; I thought mutex is simpler her since we only write to the slice in 1 place.

aldor007 (approved):

stgleb (approved):

aldor007:
LGTM

PR #140, opened 2024-10-02
fix: auto approve csr - watch ignore contex.Canceled and clear functi…

aldor007:
…on stack

From cluster-hub logs we sometimes gets
```
"retrying: fail to open v1 and v1beta watching client: fail to open v1 and v1beta watching client: context canceled" RunAutoApprove=auto-approve-csr
```

and also I think we should spin a new goroutin in case channel is close to remove the call stack of prev function

### DIFF
--- a/csr/csr.go
+++ b/csr/csr.go
@@ -281,6 +281,9 @@ func WatchCastAINodeCSRs(ctx context.Context, log logrus.FieldLogger, client kub
 		waitext.Forever,
 		func(ctx context.Context) (bool, error) {
 			w, err = getWatcher(ctx, client)
+			if errors.Is(err, context.Canceled) {
+				return false, err
+			}
 			if err != nil {
 				return true, fmt.Errorf("fail to open v1 and v1beta watching client: %w", err)
 			}
@@ -306,7 +309,8 @@ func WatchCastAINodeCSRs(ctx context.Context, log logrus.FieldLogger, client kub
 		case event, ok := <-w.ResultChan():
 			if !ok {
 				log.Debug("watcher closed")
-				WatchCastAINodeCSRs(ctx, log, client, c) // start over in case of any error
+				go WatchCastAINodeCSRs(ctx, log, client, c) // start over in case of any error
+				return
 			}
 
 			csrResult, name, request := toCertificate(event)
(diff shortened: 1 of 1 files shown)


csr/csr.go
```diff
@@ -281,6 +281,9 @@ func WatchCastAINodeCSRs(ctx context.Context, log logrus.FieldLogger, client kub
 		waitext.Forever,
 		func(ctx context.Context) (bool, error) {
 			w, err = getWatcher(ctx, client)
+			if errors.Is(err, context.Canceled) {
```
stgleb, line 284: maybe add context.DeadlineExceed as well?

stgleb (approved):
LGTM

csr/csr.go
```diff
@@ -306,7 +309,8 @@ func WatchCastAINodeCSRs(ctx context.Context, log logrus.FieldLogger, client kub
 		case event, ok := <-w.ResultChan():
 			if !ok {
 				log.Debug("watcher closed")
-				WatchCastAINodeCSRs(ctx, log, client, c) // start over in case of any error
+				go WatchCastAINodeCSRs(ctx, log, client, c) // start over in case of any error
```
ValyaB, line 312: If the watcher was closed because of canceled context, why do we start over with canceled context

csr/csr.go
ValyaB: instead of checking type of error we can check context error:
if err!=nil && ctx.Err!=nil { return false,err}

csr/csr.go
```diff
@@ -281,6 +281,9 @@ func WatchCastAINodeCSRs(ctx context.Context, log logrus.FieldLogger, client kub
 		waitext.Forever,
 		func(ctx context.Context) (bool, error) {
 			w, err = getWatcher(ctx, client)
+			if errors.Is(err, context.Canceled) {
+				return false, err
+			}
 			if err != nil {
 				return true, fmt.Errorf("fail to open v1 and v1beta watching client: %w", err)
```
ValyaB, line 288: fix error wrapping:
`return true, fmt.Errorf("getWatcher: %w", err)`

csr/csr.go
aldor007: I don't know use case for `context.DeadlineExceed` there is for contex.Canceled. 
To be safe I would like to leave it as it is.

ValyaB (approved):

PR #143, opened 2024-10-16
[create-event] fix racing

ValyaB:


furkhat (changes requested):
Please share the race error, crash output with stack trace or the failed test output.

stgleb (approved):
LGTM, just not agree with `race`.

PR #145, opened 2024-10-18
KUBE-566: send telemetry to mothership

ValyaB:


internal/monitor/metadata.go
```diff
@@ -0,0 +1,104 @@
...
+	return nil
+}
+
+// watchForMetadataChanges starts a watch on a local file for updates and returns changes to metadata channel. watcher stops when context is done
+func watchForMetadataChanges(ctx context.Context, log logrus.FieldLogger, metadataFilePath string) (chan Metadata, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("setting up new watcher: %w", err)
+	}
+	updates := make(chan Metadata, 10)
```
aldor007, line 51: Why it is a buffering channel?

internal/config/config.go
```diff
@@ -17,8 +25,15 @@ type Config struct {
 	ClusterID      string
 	PprofPort      int
 	LeaderElection LeaderElection
-	PodName        string
-	NodeName       string
+
+	MonitorMetadata string `mapstructure:"monitor_metadata"`
```
aldor007, line 29:
```suggestion
	MonitorMetadataPath string `mapstructure:"monitor_metadata"`
```

As far I can see this is path to the file but name is a bit confusing.

internal/monitor/monitor.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return fmt.Errorf("setting up metadata watch: %w", err)
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case metadata := <-metadataUpdates:
+			clusterIDHandler(metadata.ClusterID)
```
aldor007, line 35: Why do we need to update clusterID base on metadata file changes. ClusterID should be passed via env on the start and shouldn't change

ValyaB: Let me clean this logic, which left after the agent

stgleb (approved):
LGTM

internal/monitor/metadata.go
ValyaB: added non blocking write to channel and tests

aldor007 (approved):

PR #153, opened 2024-10-24
set minimum pod worker size to 1

ValyaB:


internal/actions/kubernetes_helpers.go
```diff
@@ -140,7 +140,7 @@ func executeBatchPodActions(
 	}
 
 	var (
-		parallelTasks      = int(lo.Clamp(float64(len(pods)), 30, 100))
+		parallelTasks      = lo.Clamp(len(pods), 1, 50)
```
aldor007, line 143: Why this change?

ValyaB: or when you delete a lot of empty nodes

Tsonov: But workers don't do anything, they immediately end. How can a routine doing nothing cause CPU spikes? 

Change itself is OK, although I would keep upper limit at 100.

ValyaB: We created them, so it is a spike. You end them after closing the channel.

Setting the minimum from 30 to 1 to prevent spawning goroutines more than len(pod), how can it harm us, or what are the downsides?

ValyaB: max 100 - my concern is: deleting 1k nodes -> 100k goroutines
50 doesn't fix the rate limit problem but decreases the spike.

I can leave it as is if you insist, and we can think about introducing more harder limiting with worker queue for all handlers instead of workers per handler

Tsonov (approved):

PR #154, opened 2024-10-29
KUBE-644: use informers to watch CSRs

Trojan295:


internal/actions/csr/csr.go
```diff
@@ -66,6 +68,32 @@ func (c *Certificate) Approved() bool {
 	return false
 }
 
+func (c *Certificate) Outdated() bool {
```
aldor007, line 71: Can you add a comment when certificate could become outdated?

internal/actions/csr/csr.go
```diff
@@ -273,144 +301,117 @@ func getNodeCSRV1Beta1(ctx context.Context, client kubernetes.Interface, nodeNam
...
-
-			if cert.Approved() {
-				continue
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			if err := processCSREvent(ctx, c, newObj); err != nil {
+				log.WithError(err).Warn("failed to process csr update event")
 			}
+		},
+		DeleteFunc: func(obj interface{}) {},
```
aldor007, line 348: Can this method be ommited?

internal/actions/csr/csr.go
```diff
@@ -273,144 +301,117 @@ func getNodeCSRV1Beta1(ctx context.Context, client kubernetes.Interface, nodeNam
...
+	handlerFuncs := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			if err := processCSREvent(ctx, c, obj); err != nil {
+				log.WithError(err).Warn("failed to process csr add event")
 			}
-
-			if cert.Approved() {
-				continue
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
```
aldor007, line 343: Do we need both add and updated?

internal/actions/csr/csr.go
```diff
@@ -17,16 +17,18 @@ import (
...
-	"github.com/castai/cluster-controller/internal/waitext"
+	"k8s.io/client-go/tools/cache"
 )
 
 const (
 	ReasonApproved  = "AutoApproved"
 	approvedMessage = "This CSR was approved by CAST AI"
 	csrTTL          = time.Hour
+
+	// We should approve CSRs, when they are created, so resync can be high.
```
aldor007, line 30: Can you expand above comment with information with is the resync period?

internal/actions/csr/svc.go
```diff
@@ -110,7 +110,16 @@ func (h *ApprovalManager) runAutoApproveForCastAINodes(ctx context.Context) {
 
 	log := h.log.WithField("RunAutoApprove", "auto-approve-csr")
 	c := make(chan *Certificate, 1)
-	go WatchCastAINodeCSRs(ctx, log, h.clientset, c)
+	go func() {
+		for {
+			if err := WatchCastAINodeCSRs(ctx, log, h.clientset, c); err != nil {
```
furkhat, line 115: why infinite loop here? If watch broke informer will handle reconnect. Did you think of other reasons?

internal/actions/csr/csr.go
```diff
@@ -273,144 +301,117 @@ func getNodeCSRV1Beta1(ctx context.Context, client kubernetes.Interface, nodeNam
...
+		errv1      error
+		errv1beta1 error
 	)
-	if err != nil {
-		log.Warnf("finished: %v", err)
-		return
-	}
 
-	defer w.Stop()
+	if _, errv1 = client.CertificatesV1().CertificateSigningRequests().List(ctx, metav1.ListOptions{}); errv1 == nil {
```
furkhat, line 310: I thought the idea is to use https://github.com/kubernetes/client-go/blob/master/tools/watch/informerwatcher.go

Trojan295: The `List` is to detect, if the `v1` or `v1beta` API group should be used. We could also use the Discovery API, but this would require more permissions in Kubernetes.

Didn't knew about `IndexerInformerWatcher`, is there any reason to use it instead of the `SharedInformerFactory`?

internal/actions/csr/svc.go
Trojan295: You are right, the informers handle the reconnect logic, but the `createInformer` and `informer.AddEventHandler` might fail. In such case we would end up with a running cluster-controller, but without the auto approval working.

If the context is done, then no error is returned and this goroutine quits.

internal/actions/csr/csr.go
Trojan295: Good question, in theory reacting just on `add` should be enough. Will try to remove the update and see how it behaves.

internal/actions/csr/svc.go
furkhat: Can we initialize informer in main func and pass it here? Because if informer can't be created then there is something wrong

internal/actions/csr/csr.go
furkhat: No, your way works.

internal/actions/csr/svc.go
Trojan295: Should be possible, sounds like a good idea

aldor007 (approved):

furkhat (approved):

PR #162, opened 2025-01-30
feat: Add EvictPod action

pijusn-cast:
A new action. It has some overlap with the current DrainNode action. I figured it should be OK given the small scope of this action (plus the change has no accidental impact on DrainNode action) but if you prefer to dry this up now, that's OK.

### DIFF
--- a/internal/actions/drain_node_handler_test.go
+++ b/internal/actions/drain_node_handler_test.go
@@ -514,9 +514,8 @@ func prependEvictionReaction(t testing.TB, c *fake.Clientset, success, retryable
 			if retryableFailure {
 				// Simulate failure that should be retried by client.
 				return &apierrors.StatusError{ErrStatus: metav1.Status{Reason: metav1.StatusReasonTooManyRequests}}
-			} else {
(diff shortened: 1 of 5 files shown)


internal/actions/drain_node_handler_test.go
```diff
@@ -509,29 +508,21 @@ func TestLogCastPodsToEvict(t *testing.T) {
...
-		// Simulate failure that should be retried by client.
-		if retryableFailure {
-			return true, nil, &apierrors.StatusError{ErrStatus: metav1.Status{Reason: metav1.StatusReasonTooManyRequests}}
+func prependEvictionReaction(t testing.TB, c *fake.Clientset, success, retryableFailure bool) {
+	prependPodEvictionReaction(c, func(namespace, name string) error {
+		if !success {
+			if retryableFailure {
+				// Simulate failure that should be retried by client.
+				return &apierrors.StatusError{ErrStatus: metav1.Status{Reason: metav1.StatusReasonTooManyRequests}}
+			} else {
```
andrejatcastai, line 517: NITPICK: I suggest to get rid of this `else` (generally speaking they are ugly ;)

andrejatcastai (approved):

aldor007 (approved):

PR #164, opened 2025-02-12
feat: max actions in progress

aldor007:
Part of `KUBE-417`

The idea behind the change - safeguard on the CC to limit the amount of goroutines in progress. Currently there is no limit and CC will spawn whatever API will return.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -18,15 +18,15 @@ import (
 )
 
 type Config struct {
-	Log                  Log
-	API                  API
-	TLS                  TLS
-	Kubeconfig           string
-	KubeClient           KubeClient
-	ClusterID            string
-	PprofPort            int
-	LeaderElection       LeaderElection
-        // MaxActinsInProgress serves as a safeguard to limit the number of Goroutines in progress.
+	Log            Log
+	API            API
+	TLS            TLS
+	Kubeconfig     string
+	KubeClient     KubeClient
+	ClusterID      string
+	PprofPort      int
+	LeaderElection LeaderElection
+	// MaxActionsInProgress serves as a safeguard to limit the number of Goroutines in progress.
 	MaxActionsInProgress int
 
 	MonitorMetadataPath string `mapstructure:"monitor_metadata"`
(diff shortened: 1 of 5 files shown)


internal/controller/controller.go
```diff
@@ -193,6 +194,11 @@ func (s *Controller) startProcessing(actionID string) bool {
 		return false
 	}
 
+	if inProgress := len(s.startedActions); inProgress >= s.cfg.MaxActionsInProgress {
+		s.log.Warnf("too many actions in progress %d/%d", inProgress, s.cfg.MaxActionsInProgress)
+		return false
+	}
+
```
furkhat, line 201: Interesting. So the plan is to keep pulling and then check if an action can be processed.

furkhat: Is that because `PollClusterActions` returns fixed number all the time?

furkhat: I wonder if can adjust actions API and instead of silencing actions when MaxActionsInProgress exceeded try pulling less.

aldor007: Yes.`PollClusterActions` returns actions that are not completed yet. Current max is 300. CC is during the deduplication. 

There is a plan to do a redesing of the whole API but it will take more time

internal/config/config.go
```diff
@@ -17,14 +18,15 @@ import (
...
-	LeaderElection LeaderElection
+	Log                  Log
+	API                  API
+	TLS                  TLS
+	Kubeconfig           string
+	KubeClient           KubeClient
+	ClusterID            string
+	PprofPort            int
+	LeaderElection       LeaderElection
+	MaxActionsInProgress int
```
furkhat, line 29:
```suggestion
        // MaxActinsInProgress serves as a safeguard to limit the number of Goroutines in progress.
	MaxActionsInProgress int
```

furkhat (approved):

cirisked (approved):

Tsonov:
QQ: Doesn't this open also the door for _more_ actions in parallel?

So the goal is to allow throttling on the cluster controller side, right? But adjusting the value is a manual step which customer has to do, not sure if that is the best UX. Can we detect and throttle ourselves without the var?

Tsonov (approved):

PR #168, opened 2025-02-18
feat: log name of the pods that failed to evict

aldor007:
Make it easier to debug


internal/actions/drain_node_handler.go
```diff
@@ -388,8 +388,15 @@ func (h *DrainNodeHandler) waitNodePodsTerminated(ctx context.Context, log logru
 					}
 				}
 			}
+			remainingPodsList := lo.Filter(pods, func(p v1.Pod, _ int) bool {
```
Tsonov, line 391: Suggested rewrite to combine with above and make it cleaner overall: 
- Use `Map` first to convert pods to `namespace/pod` format
- Call filter on result
- Use the final result to determine `remainingPods` + log the names if needed.

Tsonov (approved):

PR #173, opened 2025-03-20
KUBE-996: approve CSR from "kubernetes.io/kubelet-serving" for Cast nodes

ValyaB:


internal/actions/csr/csr.go
```diff
@@ -98,6 +100,11 @@ func (c *Certificate) NodeBootstrap() bool {
 	return c.RequestingUser == "kubelet-bootstrap" || c.RequestingUser == "system:serviceaccount:castai-agent:castai-cluster-controller"
 }
 
+func (c *Certificate) SystemNode() bool {
+	// To avoid waiting for the certificate to be approved by control plane.
+	return strings.HasPrefix(c.RequestingUser, "system:node:")
```
Tsonov, line 105: Doesn't this match any self-requested CSR by a node?  Including CSRs for renewing node client certificate + the CSRs for the kubelet's serving certificate. 

What does `SystemNode` mean in this context?

Tsonov:
If we are approving kubelet serving certificates, we probably should match k8s recommendation - see [docs](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-tls-bootstrapping/#client-and-serving-certificates) that offer 3 conditions for valid serving certificates

internal/actions/csr/csr.go
```diff
@@ -399,7 +422,9 @@ func sendCertificate(ctx context.Context, c chan<- *Certificate, cert *Certifica
 }
 
 func getSubjectCommonName(csrName string, csrRequest []byte) (string, error) {
-	if !strings.HasPrefix(csrName, "node-csr") {
+	// node-csr prefix for bootstrap kubelet csr.
```
Tsonov, line 425: Is this documented or empirical finding? Can we avoid relying on the name? 🤔

internal/actions/csr/csr.go
```diff
@@ -37,10 +37,12 @@ var ErrNodeCertificateNotFound = errors.New("node certificate not found")
...
 // Certificate wraps v1 and v1beta1 csr.
 type Certificate struct {
-	v1             *certv1.CertificateSigningRequest
-	v1Beta1        *certv1beta1.CertificateSigningRequest
-	Name           string
-	RequestingUser string
+	v1              *certv1.CertificateSigningRequest
+	v1Beta1         *certv1beta1.CertificateSigningRequest
+	Name            string
+	OriginalCSRName string
```
Tsonov, line 43: Why do we need `OriginalCSRName` and `Name`?

internal/actions/csr/csr.go
ValyaB: it is the username of CSR, and the idea is to approve CSRs generated by CAST nodes

internal/actions/csr/csr.go
ValyaB: no, we need the node name as a unique key to prevent the start of a new handler on the next event from the csr informer. Otherwise, we will spawn handlers in a loop.

internal/actions/csr/csr.go
ValyaB: For logging. It is hard to understand what we are signing due to rewriting 
CSR

internal/actions/csr/csr.go
Tsonov: Calling the method `IsRequestedByKubelet` or something similar would be more descriptive on what it's trying to accomplish. 
Also, at least I don't understand how the comment is related to the fact we are checking the requesting user to be a node kubelet.

internal/actions/csr/csr.go
Tsonov: I mean, this code has 2 confusing items

- Why does it return nothing for CSRs that don't match some prefix?
- The actual prefixes, are they found from a real cluster, are they generated by us? If I have to change this code in 6m, how do I know where to get prefix values from?

internal/actions/csr/csr.go
Tsonov: Isn't this data already available under `v1` or `v1beta1` ? 

I think this is where we rewrite the name, is that correct? 

```
func toCertificate(obj interface{}) (cert *Certificate, err error) {
	var name string
	var request []byte

	switch e := obj.(type) {
	case *certv1.CertificateSigningRequest:
		name = e.Name
		request = e.Spec.Request

/// <--- Why are we populating name here at all?
		cert = &Certificate{Name: name, v1: e, RequestingUser: e.Spec.Username}
	case *certv1beta1.CertificateSigningRequest:
		name = e.Name
		request = e.Spec.Request
		cert = &Certificate{Name: name, v1Beta1: e, RequestingUser: e.Spec.Username}
	default:
		return nil, errUnexpectedObjectType
	}

	cn, err := getSubjectCommonName(name, request)
	if err != nil {
		return nil, fmt.Errorf("getSubjectCommonName: Name: %v RequestingUser: %v  request: %v %w", cert.Name, cert.RequestingUser, string(request), err)
	}

/// <- here
	cert.Name = cn

	return cert, nil
}
```

internal/actions/csr/csr.go
ValyaB: added error instead of empty node name

internal/actions/csr/csr.go
ValyaB: Yes, it is avaible, but to get it, I need to check both fields. Why do you think it is not good to have an additional field?

internal/actions/csr/csr.go
ValyaB: But need to test due to it being old logic

ValyaB:
> If we are approving kubelet serving certificates, we probably should match k8s recommendation - see [docs](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-tls-bootstrapping/#client-and-serving-certificates) that offer 3 conditions for valid serving certificates

internal/actions/csr/csr.go
ValyaB: tested

internal/actions/approve_csr_handler.go
```diff
@@ -2,168 +2,20 @@ package actions
...
-
-	"github.com/castai/cluster-controller/internal/actions/csr"
 	"github.com/castai/cluster-controller/internal/castai"
-	"github.com/castai/cluster-controller/internal/waitext"
 )
 
-const (
-	approveCSRTimeout = 4 * time.Minute
-)
+// TODO deprecated action
```
Tsonov, line 9: Why not remove the code entirely?

internal/actions/csr/csr.go
Tsonov: It should be a method that reads whatever CSR is available then, having it as a field forces anyone creating a `Certificate` to know and populate the field instead of hiding this complexity.

internal/actions/csr/csr.go
```diff
@@ -398,25 +358,60 @@ func sendCertificate(ctx context.Context, c chan<- *Certificate, cert *Certifica
...
+	}
+	if err := c.validateCSR(csr); err != nil {
+		return nil, fmt.Errorf("validate CSR: %w", err)
+	}
+	return csr, nil
+}
+
+func (c *Certificate) validateCSR(csr *x509.CertificateRequest) error {
+	if c.SignerName == certv1.KubeletServingSignerName {
+		if len(csr.Subject.CommonName) == 0 {
```
Tsonov: Common name must also match the requester (the node name), right?

internal/actions/csr/csr_test.go
```diff
@@ -1,60 +1,30 @@
...
-
-	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
-	if err != nil {
-		return nil, err
-	}
-
-	clientset, err := kubernetes.NewForConfig(config)
-	return clientset, err
-}
+//func getClient() (*kubernetes.Clientset, error) {
```
Tsonov, line 18: Can be deleted

internal/actions/approve_csr_handler.go
ValyaB: The EP side can send it. I want to ensure it gets approved if EP sends it.

internal/actions/csr/csr.go
ValyaB: yes it is node name

internal/actions/csr/csr.go
```diff
@@ -398,25 +358,60 @@ func sendCertificate(ctx context.Context, c chan<- *Certificate, cert *Certifica
...
+			return fmt.Errorf("%w: CSR subject common name", errInvalidCSR)
+		}
+		if len(csr.URIs) > 0 {
```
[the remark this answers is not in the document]
ValyaB: done

internal/actions/csr/csr.go
```diff
@@ -398,25 +358,60 @@ func sendCertificate(ctx context.Context, c chan<- *Certificate, cert *Certifica
...
+			return fmt.Errorf("%w: CSR subject URIs", errInvalidCSR)
+		}
+		if len(csr.EmailAddresses) > 0 {
+			return fmt.Errorf("%w: CSR subject email addresses", errInvalidCSR)
+		}
+
+		for _, u := range c.Usages {
```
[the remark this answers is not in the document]
ValyaB: done

internal/actions/csr/csr.go
ValyaB: cleaned

internal/actions/csr/csr.go
```diff
@@ -398,25 +347,134 @@ func sendCertificate(ctx context.Context, c chan<- *Certificate, cert *Certifica
...
+	if c.v1Beta1 != nil {
+		return c.v1Beta1.Name
 	}
 
-	certReq, err := parseCSR(csrRequest)
+	return ""
+}
+
+func (c *Certificate) RequestingUser() string {
+	// node-csr prefix for bootstrap kubelet csr.
```
Tsonov, line 364: Looks like copy-paste error, requesting user is either `system:node:xxx` for serving or `kubelet-bootstrap/cluster-controller` for client auth.

internal/actions/csr/csr.go
```diff
@@ -398,25 +347,134 @@ func sendCertificate(ctx context.Context, c chan<- *Certificate, cert *Certifica
...
+	}
+	if c.v1Beta1 != nil {
+		return c.v1Beta1.Spec.Username
+	}
+
+	return ""
+}
+
+func (c *Certificate) SignerName() string {
+	// node-csr prefix for bootstrap kubelet csr.
```
Tsonov, line 377: Same copy-paste?

internal/actions/csr/csr.go
```diff
@@ -398,25 +347,134 @@ func sendCertificate(ctx context.Context, c chan<- *Certificate, cert *Certifica
...
+		if c.v1Beta1.Spec.SignerName != nil {
+			return *c.v1Beta1.Spec.SignerName
+		}
+	}
+
+	return ""
+}
+
+func (c *Certificate) Usages() []string {
+	// node-csr prefix for bootstrap kubelet csr.
```
Tsonov, line 392: same

internal/actions/csr/csr.go
```diff
@@ -398,25 +347,134 @@ func sendCertificate(ctx context.Context, c chan<- *Certificate, cert *Certifica
 	}
 }
 
-func getSubjectCommonName(csrName string, csrRequest []byte) (string, error) {
-	if !strings.HasPrefix(csrName, "node-csr") {
-		return "", nil
+func (c *Certificate) OriginalCSRName() string {
+	// node-csr prefix for bootstrap kubelet csr.
```
Tsonov, line 351: copy-paste error?

internal/actions/approve_csr_handler.go
ValyaB: renamed to ApproveCSRHandlerDeprecated

ValyaB: We should process and acknowledge the action to maintain logical consistency.

ValyaB: If EP sends approve csr action, we should handle and acknowledge it. 
if we not acknowledged - CH will try to send it again or fail AddNode action

Tsonov (approved):

PR #178, opened 2025-03-26
Draft: KUBE-1036: Ack unknown actions

Tsonov:
Previously we ignored those but did not ack so they would pile up and be resent until they expire (in 24h)

### DIFF
--- a/internal/castai/types.go
+++ b/internal/castai/types.go
@@ -3,6 +3,7 @@ package castai
 import (
 	"errors"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -13,6 +14,9 @@ const (
 	LabelNodeID               = "provisioner.cast.ai/node-id"
 	LabelManagedBy            = "provisioner.cast.ai/managed-by"
 	LabelValueManagedByCASTAI = "cast.ai"
+
+	// UnknownActionType is returned by ClusterAction.GetType when the action is not recognized.
+	UnknownActionType = "Unknown"
 )
 
 type GetClusterActionsResponse struct {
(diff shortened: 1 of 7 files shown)


internal/castai/types.go
```diff
@@ -94,6 +98,21 @@ func (c *ClusterAction) Data() interface{} {
 	return nil
 }
 
+// IsValid checks if the action is OK to use.
+// If this value is nil, most likely current version of cluster controller does not know about the action type and cannot execute it.
+// It can also be a case of invalid data from server, but we cannot distinguish between the two at the moment.
+func (c *ClusterAction) IsValid() bool {
```
furkhat, line 104: IsKnown?

internal/castai/types.go
```diff
@@ -94,6 +98,21 @@ func (c *ClusterAction) Data() interface{} {
...
+	return c.Data() != nil
+}
+
+// GetType tries to deduct the type of the action based on its data. In case this fails, UnknownActionType is returned.
+func (c *ClusterAction) GetType() string {
+	if c.Data() != nil {
+		return reflect.TypeOf(c.Data()).String()
+	}
+	return UnknownActionType
+}
```
furkhat, line 114: Just thought may be it is more "Golang way" to do like this?
```suggestion
// GetType tries to deduct the type of the action based on its data. In case this fails, UnknownActionType is returned.
func (c *ClusterAction) GetType() (string, error) {
	if c.Data() != nil {
		return reflect.TypeOf(c.Data()).String(), nil
	}
	return "", ErrUnknownActionType
}
```

Tsonov: I don't think returning error here is good ergonomics. If there is no data on the action, it is either (1) invalid or (2) from a type that CC doesn't understand. CC can't do anything except reject the action in both cases.

It would force every caller of `Type` to check for error, which (in practice) it cannot handle and has to just surface. Instead, we should filter those actions early (like we do with this PR) and let the code in handlers be simpler, assuming it can use `Type()` directly without a forced error check.

That's my 2 cents

For `GetType` vs `Type` I have no preference, if you think it's more idiomatic, I'll change.

PR #179, opened 2025-03-27
NOJIRA: Remove obsolete action

Tsonov:
We now have handling to ack unknown actions with error, no need for noop.


ValyaB:
clean needed here https://github.com/castai/cluster-controller/blob/cb9cafb57632ab3485bbd7bb9a360f30d0413f7c/internal/castai/types.go#L171

ValyaB (approved):

PR #181, opened 2025-03-28
KUBE-1046: disable autoapproving csr when autoscaling disabled

ValyaB:


aldor007 (approved):

internal/config/config.go
```diff
@@ -27,6 +27,8 @@ type Config struct {
 	PprofPort      int
 	Metrics        Metrics
 	LeaderElection LeaderElection
+	// AutoscalingDisabled is a flag to disable approving csr.
+	AutoscalingDisabled bool `mapstructure:"autoscaling_disabled"`
```
Tsonov, line 31: [Nit] Helm chart uses positive notion (`AutoscalingEnabled`), why not be consistent with it?

aldor007: There could be an issue with defaut value. You need to set in values `AutoscalingEnabled: true` to be a backward compatible. Helm upgrade without one flag reset values to default for boolean false so even though in values.yaml there is true helm upgrade will set false 


So having default behavior the same is safer option

ValyaB: By default, now AutoscalingEnabled = false in the environment
Want to avoid the default behavior by disabling 
If disabled=true, then we turn it off

Tsonov: thanks for explanation, didn't know the helm oddity

PR #186, opened 2025-04-15
KUBE-1098: update packages

ValyaB:


.github/workflows/pr.yaml
```diff
@@ -75,6 +75,16 @@ jobs:
           push: true
           tags: ghcr.io/castai/cluster-controller:${{ env.VERSION_TAG }}
 
+      - name: Run Trivy vulnerability scanner
```
Tsonov, line 78: Not a blocker but doesn't FOSSA already show this ? Asked in #internal-security

Tsonov (approved):

.github/workflows/pr.yaml
ValyaB: it is not, also asked Patrick

PR #188, opened 2025-05-27
KUBE-1216: add provider id check to exclude node name reusing

ValyaB:


internal/actions/kubernetes_helpers.go
```diff
@@ -85,37 +86,46 @@ func patchNodeStatus(ctx context.Context, log logrus.FieldLogger, clientset kube
...
+	if err != nil && k8serrors.IsNotFound(err) {
+		return nil, errNodeNotFound
+	}
+	if err != nil {
+		return nil, err
+	}
 
-	var node *v1.Node
+	if n == nil {
+		return nil, nil
```
Tsonov, line 103: should we return `errNodeNotFound` here as well?

internal/actions/kubernetes_helpers.go
```diff
@@ -85,37 +86,46 @@ func patchNodeStatus(ctx context.Context, log logrus.FieldLogger, clientset kube
...
+	}
 
-	boff := waitext.DefaultExponentialBackoff()
-	boff.Duration = 5 * time.Second
+	if err := isNodeIDProviderIDValid(n, nodeID, providerID); err != nil {
+		return nil, fmt.Errorf("node %s not valid %w", n.Name, err)
```
Tsonov, line 107: [nit] I'd say the node is valid but it doesn't match what was requested (so `invalid` -> `does not match`)

internal/actions/types.go
```diff
@@ -17,7 +17,10 @@ const (
 	ActionIDLogField = "id"
 )
 
-var errAction = errors.New("not valid action")
+var (
+	errAction       = errors.New("not valid action")
+	errNodeNotFound = errors.New("node not found")
```
Tsonov, line 22: Is this duplicate from `actions`? We also have similar errors there

internal/actions/kubernetes_helpers.go
```diff
@@ -85,37 +86,46 @@ func patchNodeStatus(ctx context.Context, log logrus.FieldLogger, clientset kube
...
-		5,
-		func(ctx context.Context) (bool, error) {
-			var err error
-			node, err = clientset.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
-			if err != nil {
-				return true, err
+	return n, nil
+}
+
+var errNodeNotValid = fmt.Errorf("node is not valid")
```
Tsonov, line 113: Let's move this to `types.go` with the other errors to make it easier to find.

internal/actions/kubernetes_helpers.go
```diff
@@ -85,37 +86,46 @@ func patchNodeStatus(ctx context.Context, log logrus.FieldLogger, clientset kube
...
-		return nil, err
+		}
 	}
-	return node, nil
+
+	if providerID != "" && node.Spec.ProviderID != "" && node.Spec.ProviderID == providerID {
+		return nil
+	}
+
+	return fmt.Errorf("node ID %s or provider ID %s does not match node %s: %w", nodeID, providerID, node.Name, errNodeNotValid)
```
Tsonov, line 128: let's also log the node's providerID and nodeID values (if they exist); would make debugging a lot easier via logs

internal/actions/kubernetes_helpers_test.go
```diff
@@ -0,0 +1,121 @@
...
+						Labels: map[string]string{
+							castai.LabelNodeID: "node-id-123",
+						},
+					},
+				},
+				nodeID: "node-id-123",
+			},
+		},
+		{
+			name: "node ID does not match label but provider ID matches",
```
Tsonov, line 45: if label exists and doesn't match, I'd expect to get invalid node error?

internal/actions/check_node_deleted.go
```diff
@@ -63,33 +61,15 @@ func (h *CheckNodeDeletedHandler) Handle(ctx context.Context, action *castai.Clu
 		boff,
 		h.cfg.retries,
 		func(ctx context.Context) (bool, error) {
-			n, err := h.clientset.CoreV1().Nodes().Get(ctx, req.NodeName, metav1.GetOptions{})
-			if apierrors.IsNotFound(err) {
-				return false, nil
+			n, err := getNodeByIDs(ctx, h.clientset, req.NodeName, req.NodeID, req.ProviderId)
+			if n != nil {
```
Tsonov, line 65: [nit] In idiomatic go, i'd expect to see
```
			if errors.Is(err, errNodeNotFound) {
				return false, nil
			}

if err != nil {  // We already know it's not notFound
 return true, err
}

// Use n here as err == nil
```

Not critical though

internal/actions/check_node_deleted_test.go
```diff
@@ -0,0 +1,112 @@
...
+						Spec: v1.NodeSpec{
+							ProviderID: "providerID",
+						},
+					},
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "handle check successfully when node is not found",
```
Tsonov, line 84: Should we also test the case where ID and providerID do not match?

internal/actions/check_node_status_test.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/castai/cluster-controller/internal/castai"
 )
 
+// nolint:goconst
 func TestCheckStatus_Deleted(t *testing.T) {
```
Tsonov, line 23: I think we need a few more tests to cover nodeID/providerID permutations. See comment above about possible scenarios

internal/actions/delete_node_handler_test.go
```diff
@@ -27,6 +27,9 @@ func TestDeleteNodeHandler(t *testing.T) {
 		node := &v1.Node{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: nodeName,
```
Tsonov, line 29: Add a few tests to cover prodiverID/nodeID permutations

internal/actions/drain_node_handler.go
```diff
@@ -83,16 +83,17 @@ func (h *DrainNodeHandler) Handle(ctx context.Context, action *castai.ClusterAct
...
 		"node_name":      req.NodeName,
 		"node_id":        req.NodeID,
+		"provider_id":    req.ProviderId,
 		"action":         reflect.TypeOf(action.Data().(*castai.ActionDrainNode)).String(),
 		ActionIDLogField: action.ID,
 	})
 
-	node, err := h.clientset.CoreV1().Nodes().Get(ctx, req.NodeName, metav1.GetOptions{})
+	node, err := getNodeByIDs(ctx, h.clientset, req.NodeName, req.NodeID, req.ProviderId)
+	if errors.Is(err, errNodeNotFound) {
```
Tsonov, line 92: If we have a mismatch, shouldn't we also skip the drain?

internal/actions/delete_node_handler_test.go
ValyaB: added here Test_isNodeIDProviderIDValid

internal/actions/check_node_status.go
```diff
@@ -71,42 +72,28 @@ func (h *CheckNodeStatusHandler) checkNodeDeleted(ctx context.Context, log *logr
...
-			}
-			if currentNodeID != "" {
-				if currentNodeID != req.NodeID {
-					log.Info("node name was reused. Original node is deleted")
-					return false, nil
-				}
-				if currentNodeID == req.NodeID {
-					return false, fmt.Errorf("current node id is equal to requested node id: %v %w", req.NodeID, errNodeNotDeleted)
-				}
+			if errors.Is(err, errNodeNotValid) {
```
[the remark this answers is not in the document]
ValyaB: I would like to create separate use cases for isNodeIDProviderIDValid 
+ per action on top of isNodeIDProviderIDValid

internal/actions/drain_node_handler.go
ValyaB: We skip it with an error

internal/actions/kubernetes_helpers_test.go
```diff
@@ -0,0 +1,480 @@
...
+			name: "empty node ID and provider id in request",
+			args: args{
+				node:       &v1.Node{},
+				providerID: "",
+				nodeID:     "",
+			},
+			wantErr: errAction,
+		},
+		{
+			name: "request node ID is empty but node id exists in node labels",
```
Tsonov, line 42: if we will support an `or` for providerID vs nodeID, then in this case we should accept the match by providerID and execute the operation?

internal/actions/kubernetes_helpers.go
```diff
@@ -85,37 +87,69 @@ func patchNodeStatus(ctx context.Context, log logrus.FieldLogger, clientset kube
...
+	}
+	emptyProviderID := providerID == "" || node.Spec.ProviderID == ""
+	validProviderID := false
+
+	// validate provider id only if non-empty in request and in Node spec
+	// Azure provider: provider id can be empty even if node is Ready
+	if !emptyProviderID && node.Spec.ProviderID == providerID {
+		validProviderID = true
+	} else if !emptyProviderID && node.Spec.ProviderID != providerID {
+		log.Errorf("node %v has provider ID %s, but requested provider ID is %s", node.Name, node.Spec.ProviderID, providerID)
```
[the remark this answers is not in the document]
ValyaB: name: "node ID is match and provider ID does not match",

internal/actions/kubernetes_helpers_test.go
ValyaB: yes  here we are not expecting error, was failed testcase - fixed

internal/actions/check_node_status.go
```diff
@@ -71,80 +83,85 @@ func (h *CheckNodeStatusHandler) checkNodeDeleted(ctx context.Context, log *logr
...
-			}
+			return checkNodeDeleted(ctx, h.clientset.CoreV1().Nodes(), req.NodeName, req.NodeID, req.ProviderId, log)
+		},
+		func(err error) {
+			log.Warnf("check node %s status failed, will retry: %v", req.NodeName, err)
+		},
+	)
+}
 
-			// If node is nil - deleted
```
aldor007, line 79: Can you leave this comment?

internal/actions/kubernetes_helpers.go
```diff
@@ -85,37 +87,69 @@ func patchNodeStatus(ctx context.Context, log logrus.FieldLogger, clientset kube
...
-	// as a result was returned from watch. But subsequent get request returns not found.
-	// This is likely due to clientset's caching that's meant to alleviate API's load.
-	// So we give enough time for cache to sync - ~10s max.
+func getNodeByIDs(ctx context.Context, clientSet corev1.NodeInterface, nodeName, nodeID, providerID string, log logrus.FieldLogger) (*v1.Node, error) {
+	if nodeID == "" && providerID == "" {
+		return nil, fmt.Errorf("node and provider IDs are empty %w", errAction)
+	}
 
-	var node *v1.Node
+	n, err := clientSet.Get(ctx, nodeName, metav1.GetOptions{})
```
aldor007, line 95: Why remove Retry?

```
// on GKE we noticed that sometimes the node is not found, even though it is in the cluster
// as a result was returned from watch. But subsequent get request returns not found.
// This is likely due to clientset's caching that's meant to alleviate API's load.
// So we give enough time for cache to sync - ~10s max.
```

The comment is still valid

internal/actions/kubernetes_helpers.go
```diff
@@ -85,37 +87,69 @@ func patchNodeStatus(ctx context.Context, log logrus.FieldLogger, clientset kube
...
-				return true, err
+	if err := isNodeIDProviderIDValid(n, nodeID, providerID, log); err != nil {
+		return nil, fmt.Errorf("requested node ID %s, provider ID %s for node name: %s %w",
+			nodeID, providerID, n.Name, err)
+	}
+
+	return n, nil
+}
+
+func isNodeIDProviderIDValid(node *v1.Node, nodeID, providerID string, log logrus.FieldLogger) error {
```
aldor007, line 115: The method is very complex. It requires a proper godoc IMO

internal/actions/check_node_status.go
ValyaB: added to checkNodeDeleted due to this function being used by check_node status and dele_node handlers

internal/actions/kubernetes_helpers.go
ValyaB: All permutations are detailed in the tests. If I miss anything, let me know. Additionally, all permutations are duplicated in the tests for each handler.

aldor007: When debuging first thing is reading code.  I still insist on adding doc here.

internal/actions/drain_node_handler.go
```diff
@@ -339,6 +349,7 @@ func (h *DrainNodeHandler) listNodePodsToEvict(ctx context.Context, log logrus.F
 			continue
 		}
 
+		// skip pods created on the node with the same name.
```
Tsonov, line 352: this looks obsolete, could be removed

internal/actions/check_node_status.go
```diff
@@ -71,80 +83,89 @@ func (h *CheckNodeStatusHandler) checkNodeDeleted(ctx context.Context, log *logr
...
-			// If label is present and doesn't match - node was reused - deleted
-			// If label is present and matches - node is not deleted
-			// If label is not present and node is not nil - node is not deleted (potentially corrupted state).
+func checkNodeDeleted(ctx context.Context, clientSet v1.NodeInterface, nodeName, nodeID, providerID string, log logrus.FieldLogger) (bool, error) {
+	// If node is nil - deleted
+	// If label is present and doesn't match - node was reused - deleted
```
Tsonov, line 96: We can adjust this a bit to say that if providerID or label have mismatch, then it's reused and deleted

aldor007 (approved):

Tsonov (approved):

PR #195, opened 2025-08-05
feat: export metrics to CastAI

Trojan295:


internal/castai/client.go
```diff
@@ -169,3 +191,54 @@ func (c *Client) AckAction(ctx context.Context, actionID string, req *AckCluster
...
+					continue
+				}
+
+				timeserie.Labels = append(timeserie.Labels, PrometheusLabel{
+					Name:  *label.Name,
+					Value: lo.FromPtr(label.Value),
+				})
+			}
+
+			if metric.Counter != nil {
```
Tsonov, line 229: Isn't this redundant, after line 202 you know it's not nil

cmd/controller/run.go
```diff
@@ -55,7 +56,7 @@ func run(ctx context.Context) error {
 		log.Fatalf("failed to create castai client: %v", err)
 	}
 
-	client := castai.NewClient(logger, cl, cfg.ClusterID)
```
Tsonov, line 58: This will be empty locally, is it a problem?

cmd/controller/run.go
```diff
@@ -159,6 +160,11 @@ func runController(
 		}
 	}()
 
+	if cfg.Metrics.ExportEnabled {
+		metricExporter := metricexporter.New(log, client, cfg.Metrics.ExportInterval)
```
Tsonov, line 164: let's add a log that this is enabled so we know

Tsonov (approved):

